PR #2, opened 2023-02-03
Do not treat warnings as errors when building dynarmic

DCNick3:
Building on my (arch) linux gives some spurious warnings that get treated as errors, probably because of newer gcc (12)

Also don't build tests, probably saving a bit compilation time


hikari-no-yume:
Aha, a friend of mine already tried disabling tests and I think it was also to save compilation time. Looks good, thanks a lot!

PR #18, opened 2023-02-04
Update README.md

upintheairsheep:
iOS 6 should be the last supported version

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 As an HLE, touchHLE is radically different from a low-level emulator (LLE) like QEMU. The only code the [emulated CPU](https://github.com/merryhime/dynarmic) executes is the app binary and [a handful of libraries](touchHLE_dylibs/); touchHLE takes the place of iPhone OS and provides its own implementations of the system frameworks (Foundation, UIKit, OpenGL ES, OpenAL, etc).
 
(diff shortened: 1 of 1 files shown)


hikari-no-yume:
I'm intentionally keeping the range of versions small because I already know iOS 4.x adds some things that may make supporting it a particular challenge, let alone iOS 5 or iOS 6. At the same time, as much as I like iOS 6 being the last one with the classic æsthetics, it's not the last 32-bit version, so it's an arbitrary goal. So, sorry, I'm not going to change the README right now. It's not like the README actually determines where the project will go, in any case. That is down to the efforts put in by contributors (and for the time being, probably mostly my efforts). The function of the README is really to set expectations, and I don't want to suggest iOS 6.x is likely until I know it's actually doable.

upintheairsheep:
Thanks, also check out the Cycada resources I sent you, and remember that you could either physically mail them about the code, or just try making your own version of Cycada via the papers or let it inspire more stuff in here

upintheairsheep:
IPASim also has nearly the exact goal of yours, and copying code from it might be helpful as well. Try using that QEMU iPhone OS fork as well to help with the project, I am a big fan of this project by the way.

hikari-no-yume:
Please slow down with how often you are commenting. I do appreciate your enthusiasm but I don't want to spend all my time on this.

PR #19, opened 2023-02-04
Free apps exist, so the copyright disclaimer has to be changed to remember them

upintheairsheep:


upintheairsheep:
Sorry for my spamming, I am just so happy that an iOS emulator finally exists, not one, but TWO, the other one is a QEMU iPhone OS 2 fork

hikari-no-yume:
I think this caveat is going to be obvious so I'd rather not complicate the README with it.

I appreciate the enthusiasm, but can you slow down a bit?

Issue #33, opened 2023-02-05
Use anyhow or some other systematic error handling approach


DCNick3:
Currently, some errors are handled as just `Result<T, &'static str>` or `Result<T, String>`. While this may be fine in some cases, I suggest using a more systematic approach, like, for example, `anyhow` provides.

Once of its nicer features is providing context to previous errors, giving nicer error messages overall.

For example:

```
fn work_with_files() -> anyhow::Result {
  std::fs::read("hello.txt").context("reading hello.txt")?;
  std::fs::write("bye.txt", "sayonara").context("writing bye.txt")?;
}
```

When printing the error not only the underlying IoError will be printed, but the context chain as well, like:

```
reading hello.txt
Caused by:
- Os { code: 9, kind: Uncategorized, message: "Bad file descriptor" }
```

Maybe you should also consider `snafu`: it gives similar features (context too), but allows you to type all your errors (still providing the `Whatever` escape hatch). This might be useful for exposing errors to the guest, as they are expected to be classified as some `errno` code, which is not easy for pure string errors...

hikari-no-yume:
Mm, optimal error handling is a topic I'm not super familiar with in Rust. I know how to use Result and Option pretty well, but I haven't even touched Error, let alone third-party libraries.

On that note, in general I want to keep the number of dependencies pretty small and therefore I want to avoid using an external dependency unless I really have to, so if there's a way to do something with just what `std` provides, even if it's a bit more fiddly or verbose, I prefer to do it that way. I think I picked up this habit from C programming, and I know it's not as necessary in Rust, but considering the sheer number of things an iPhone OS emulator is going to need to do (audio, video, graphics, text, input, and so much more), I think it'll be good for keeping the binary size and build time under control.

Anyway, I have some thoughts about error management in this project in general.

The overwhelmingly most common error handling approach is to panic, directly or indirectly (`unwrap()`, `assert!`, etc). I believe that's generally discouraged as a last resort, but in this project, most of the time there's no better alternative. Usually the error is related to the emulator not supporting something the app needs, or the app doing something unexpected, and in either case it's safest to immediately crash so it doesn't continue into a broken state. Perhaps one reason people don't like panics, especially unwraps, is they don't necessarily give good context, but I have a nice panic handler that gives you a backtrace for the guest code (the emulated stack, which exists in parallel to the host/Rust stack), and I think that's often the best information we possibly could offer.

[...]

PR #34, opened 2023-02-05
Allow opening an IPA bundle file directly

### LINKED ISSUE #20: Add support for opening IPA files
Please add support for opening IPA files directly, using the 7zip or a normal zip engine to extract it and run it.

DCNick3:
Closes #20

It works, but I am not yet happy about all the code. 
The error handling is kind of all over the place.
Also, maybe it's worth making a `FileRef` enum representing "a file that can potentially be opened" instead of two variants in `FsNode` enum

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c71b1793ee61086797f5c80b6efa2b8ffa6d5dd703f118545808a7f2e27f7046"
 
 [[package]]
+name = "adler"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+
+[[package]]
 name = "ansi_term"
 version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -192,6 +198,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "crc32fast"
+version = "1.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b540bd8bc810d3885c6ea91e2018302f68baba2129ab3e88f32389ee9370880d"
+dependencies = [
+ "cfg-if",
+]
+
+[[package]]
+name = "crossbeam-utils"
+version = "0.8.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4fb766fa798726286dbbb842f174001dab8abc7b627a1dd86e0b7222a95d929f"
+dependencies = [
+ "cfg-if",
+]
+
+[[package]]
 name = "csv"
 version = "1.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -211,6 +235,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2b2466559f260f48ad25fe6317b3c8dac77b5bdb5763ac7d9d6103530663bc90"
 dependencies = [
  "memchr",
+]
+
+[[package]]
+name = "flate2"
+version = "1.0.25"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a8a2db397cb1c8772f31494cb8917e48cd1e64f0fa7efac59fbd741a0a8ce841"
+dependencies = [
+ "crc32fast",
+ "miniz_oxide",
 ]
 
 [[package]]
@@ -339,6 +373,15 @@ name = "memchr"
 version = "2.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2dffe52ecf27772e601905b7522cb4ef790d2cc203488bbd0e2fe85fcb74566d"
+
+[[package]]
+name = "miniz_oxide"
+version = "0.6.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b275950c28b37e794e8c55d88aeb5e139d0ce23fdbbeda68f8d7174abdf9e8fa"
+dependencies = [
+ "adler",
+]
 
 [[package]]
 name = "once_cell"
@@ -616,6 +659,7 @@ dependencies = [
  "touchHLE_gl_bindings",
  "touchHLE_openal_soft_wrapper",
  "touchHLE_stb_image_wrapper",
+ "zip",
 ]
 
 [[package]]
@@ -719,3 +763,15 @@ name = "xml-rs"
 version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d2d7d3948613f75c98fd9328cfdcc45acc4d360655289d0a7d4ec931392200a3"
+
+[[package]]
+name = "zip"
+version = "0.6.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0445d0fbc924bb93539b4316c11afb121ea39296f99a3c4c9edad09e3658cdef"
+dependencies = [
+ "byteorder",
+ "crc32fast",
+ "crossbeam-utils",
+ "flate2",
+]
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -20,14 +20,24 @@
 //! Directories only need a corresponding directory in the host filesystem if
 //! they are writeable (i.e. if new files can be created in them).
 
+mod bundle;
+
+pub use bundle::BundleData;
+
+use crate::fs::bundle::{IpaFile, IpaFileRef};
 use std::collections::HashMap;
+use std::fs::File;
+use std::io::{Read, Seek, Write};
 use std::path::{Path, PathBuf};
 
 #[derive(Debug)]
 enum FsNode {
-    File {
+    HostFile {
         host_path: PathBuf,
         writeable: bool,
+    },
+    IpaBundleFile {
+        file: IpaFileRef,
     },
     Directory {
         children: HashMap<String, FsNode>,
@@ -55,7 +65,7 @@ impl FsNode {
             if kind.is_file() {
                 children.insert(
                     name,
-                    FsNode::File {
+                    FsNode::HostFile {
                         host_path,
                         writeable,
                     },
@@ -91,11 +101,14 @@ impl FsNode {
         assert!(children.insert(String::from(name), child).is_none());
         self
     }
-    fn file(host_path: PathBuf) -> Self {
-        FsNode::File {
+    fn host_file(host_path: PathBuf) -> Self {
+        FsNode::HostFile {
             host_path,
             writeable: false,
         }
+    }
+    fn bundle_zip_file(file: IpaFileRef) -> Self {
+        FsNode::IpaBundleFile { file }
     }
 }
 
@@ -104,7 +117,7 @@ impl FsNode {
 #[derive(Debug)]
 pub struct GuestPath(str);
 impl GuestPath {
-    pub fn new<S: AsRef<str>>(s: &S) -> &GuestPath {
+    pub fn new<S: AsRef<str> + ?Sized>(s: &S) -> &GuestPath {
         unsafe { &*(s.as_ref() as *const str as *const GuestPath) }
     }
     pub fn as_str(&self) -> &str {
@@ -120,11 +133,23 @@ impl GuestPath {
         GuestPathBuf::from(format!("{}/{}", self.as_str(), path.as_ref()))
     }
 
+    /// Splits the path into a parent path and a file name.
+    pub fn parent_and_file_name(&self) -> Option<(&GuestPath, &str)> {
+        // FIXME: this should do the same resolution as `std::path::file_name()`
+        let (parent_name, file_name) = self.as_str().rsplit_once('/')?;
+        Some((GuestPath::new(parent_name), file_name))
+    }
+
     /// Get the final component of the path.
     pub fn file_name(&self) -> Option<&str> {
-        // FIXME: this should do the same resolution as `std::path::file_name()`
-        let (_, file_name) = self.as_str().rsplit_once('/')?;
+        let (_, file_name) = self.parent_and_file_name()?;
         Some(file_name)
+    }
+
+    /// Get the parent directory of the path.
+    pub fn parent(&self) -> Option<&GuestPath> {
+        let (parent_name, _) = self.parent_and_file_name()?;
+        Some(parent_name)
     }
 }
 impl AsRef<GuestPath> for GuestPath {
@@ -142,7 +167,7 @@ impl AsRef<GuestPath> for str {
         unsafe { &*(self as *const str as *const GuestPath) }
     }
 }
-impl std::borrow::ToOwned for GuestPath {
+impl ToOwned for GuestPath {
     type Owned = GuestPathBuf;
 
     fn to_owned(&self) -> GuestPathBuf {
@@ -276,6 +301,68 @@ fn handle_open_err<T>(open_result: std::io::Result<T>, host_path: &Path) -> T {
     }
 }
 
+/// Like [File] but for the guest filesystem.
+#[derive(Debug)]
+pub enum GuestFile {
+    HostFile(File),
+    IpaBundleFile(IpaFile),
+}
+
+impl GuestFile {
+    fn from_host_file(file: File) -> GuestFile {
+        GuestFile::HostFile(file)
+    }
+
+    fn from_ipa_file(file: &IpaFileRef) -> GuestFile {
+        GuestFile::IpaBundleFile(file.open())
+    }
+
+    pub fn sync_all(&self) -> std::io::Result<()> {
+        match self {
+            GuestFile::HostFile(file) => file.sync_all(),
+            GuestFile::IpaBundleFile(_) => Ok(()),
+        }
+    }
+}
+
+impl Read for GuestFile {
+    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
+        match self {
+            GuestFile::HostFile(file) => file.read(buf),
+            GuestFile::IpaBundleFile(file) => file.read(buf),
+        }
+    }
+}
+
+impl Write for GuestFile {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        match self {
+            GuestFile::HostFile(file) => file.write(buf),
+            GuestFile::IpaBundleFile(file) => {
+                panic!("Attempt to write to a read-only file: {:?}", file)
+            }
+        }
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        match self {
+            GuestFile::HostFile(file) => file.flush(),
+            GuestFile::IpaBundleFile(file) => {
+                panic!("Attempt to flush a read-only file: {:?}", file)
+            }
+        }
+    }
+}
+
+impl Seek for GuestFile {
+    fn seek(&mut self, pos: std::io::SeekFrom) -> std::io::Result<u64> {
+        match self {
+            GuestFile::HostFile(file) => file.seek(pos),
+            GuestFile::IpaBundleFile(file) => file.seek(pos),
+        }
+    }
+}
+
 /// The type that owns the guest filesystem and provides accessors for it.
 #[derive(Debug)]
 pub struct Fs {
@@ -299,7 +386,7 @@ impl Fs {
     /// sandbox directory, where documents can be stored. A directory will be
     /// created at that path if it does not already exist.
     pub fn new(
-        bundle_host_path: &Path,
+        app_bundle: BundleData,
         bundle_dir_name: String,
         bundle_id: &str,
     ) -> (Fs, GuestPathBuf) {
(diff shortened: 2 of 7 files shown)


Cargo.lock
```diff
@@ -191,6 +266,49 @@ dependencies = [
...
+name = "crossbeam-utils"
+version = "0.8.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4fb766fa798726286dbbb842f174001dab8abc7b627a1dd86e0b7222a95d929f"
+dependencies = [
+ "cfg-if",
+]
+
+[[package]]
+name = "crypto-common"
```
hikari-no-yume: Hmm, I assume this is being pulled in because of default features in the `zip` crate, presumably to support ZIP files with encryption, but I don't think we need it for IPA files? There's a few other new dependencies here that are presumably also not needed. Can you try cutting down the feature list for this dependency to just an empty one?

src/fs.rs
```diff
@@ -91,20 +100,23 @@ impl FsNode {
...
+    }
 }
 
 /// Like [Path] but for the virtual filesystem.
 #[repr(transparent)]
 #[derive(Debug)]
 pub struct GuestPath(str);
 impl GuestPath {
-    pub fn new<S: AsRef<str>>(s: &S) -> &GuestPath {
+    pub fn new<S: AsRef<str> + ?Sized>(s: &S) -> &GuestPath {
```
hikari-no-yume, line 120: Oh, this is surprising to me because I thought this code worked. Was this constructor not working because I failed to include this bound? This looks fine though, I'm just wondering.

src/fs.rs
```diff
@@ -276,6 +300,64 @@ fn handle_open_err<T>(open_result: std::io::Result<T>, host_path: &Path) -> T {
...
+}
+
+impl File {
+    fn from_host_file(file: std::fs::File) -> File {
+        File::HostFile(file)
+    }
+
+    fn from_ipa_file(file: IpaFile) -> File {
+        File::IpaBundleFile(file)
+    }
```
hikari-no-yume: A bit of a nitpick I guess, but since these aren't public, maybe we should just create the enum members directly for the time being? The only reason I have constructor functions for the `FsNode`s is that they helped a lot with the readability in `Fs::new`.

Actually, I notice that all the uses of `from_ipa_file` are paired with `file.open()` right now, so perhaps you could move that bit into this constructor. Then it would be justifying its own existence.

src/fs.rs
```diff
@@ -276,6 +300,64 @@ fn handle_open_err<T>(open_result: std::io::Result<T>, host_path: &Path) -> T {
     }
 }
 
+/// Like [std::fs::File] but for the guest filesystem.
+#[derive(Debug)]
+pub enum File {
```
hikari-no-yume: Could you rename this to `GuestFile`? I try to avoid using the same names as well-known `std` types, especially if they're conceptually very similar, so that they'll be easy to distinguish if we import both in the same file. Hence `GuestPath`, `GuestUSize`, etc.

If you do, you can then add `use std::fs::File;` at the top of this file, if you feel like it.

src/fs.rs
```diff
@@ -418,40 +494,40 @@ impl Fs {
 
     /// Like [std::path::Path::is_file] but for the guest filesystem.
     pub fn is_file(&self, path: &GuestPath) -> bool {
-        matches!(self.lookup_node(path), Some(FsNode::File { .. }))
+        matches!(self.lookup_node(path), Some(FsNode::HostFile { .. }))
```
hikari-no-yume: Uhh, this looks like a bug, we want files from the IPA bundle to be considered “files” too, right?

.gitignore
```diff
@@ -1,3 +1,4 @@
 /target
 /vendor/boost
 /touchHLE_sandbox
+.idea
```
hikari-no-yume: Was this added automatically by your IDE? I think there's some way to tell git to ignore files locally, without having to modify the repo.

src/fs.rs
```diff
@@ -276,6 +300,64 @@ fn handle_open_err<T>(open_result: std::io::Result<T>, host_path: &Path) -> T {
...
+            File::IpaBundleFile(file) => file.read(buf),
+        }
+    }
+}
+
+impl Write for File {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        match self {
+            File::HostFile(file) => file.write(buf),
+            File::IpaBundleFile(file) => file.write(buf),
```
hikari-no-yume: I think we'll always want to treat the IPA file as read-only. I see you have a `Write` implementation in `src/fs/bundle.rs`, but maybe you could remove it and just panic here instead.

src/fs.rs
```diff
@@ -418,40 +494,40 @@ impl Fs {
...
+        let mut result = Vec::new();
+        file.read_to_end(&mut result).map_err(|_| ())?;
+        Ok(result)
     }
 
     /// Like [std::fs::File::open] but for the guest filesystem.
     #[allow(dead_code)]
-    pub fn open<P: AsRef<GuestPath>>(&self, path: P) -> Result<std::fs::File, ()> {
+    pub fn open<P: AsRef<GuestPath>>(&self, path: P) -> Result<File, ()> {
+        // it would be nice to delegate to self.open_with_options, but currently it wants a mutable reference to self
```
hikari-no-yume, line 519: Ah, so you ran into this too, that's indeed exactly why I have the two implementations. I'm a bit scared they'll diverge in behaviour in some corner case eventually, but not requiring `&mut` is nice. Maybe I should write some tests for this code, it's some of the trickiest stuff to get right and also the easiest to test.

(Just dismiss this, I'm not asking you to change the comment, it's good!)

src/fs/bundle.rs
```diff
@@ -0,0 +1,235 @@
...
+}
+
+impl FsNodeBuilder {
+    fn new() -> Self {
+        Self {
+            root: FsNode::dir(),
+        }
+    }
+
+    fn find_or_make_directory(&mut self, path: &GuestPath) -> &mut FsNode {
```
hikari-no-yume, line 29: Just curious, is this because the entry for `/SomeDir/SomeFoo` can appear before the entry for `/SomeDir` or something like that?

src/fs/bundle.rs
```diff
@@ -0,0 +1,235 @@
...
+            let next = children
+                .entry(part.to_string())
+                .or_insert_with(|| FsNode::dir());
+            current = next;
+        }
+        current
+    }
+
+    fn add_file(&mut self, path: &GuestPath, node: FsNode) {
+        // I __hope__ no adversory will try to create files with .. in their names...
```
hikari-no-yume: Maybe `assert!` just in case. (FWIW I think the right spelling is `adversary`.)

src/fs/bundle.rs
```diff
@@ -0,0 +1,235 @@
...
+    Zip {
+        zip: ZipArchive<std::fs::File>,
+        /// Path to the app bundle inside the zip file.
+        /// It should be "Payload/<app name>.app" (no trailing slash!).
+        bundle_path: String,
+    },
+}
+
+impl BundleData {
+    // TODO actually, the error type is not at all a ZipError.. I would like to use a custom error type here (anyhow, snafu, thiserror, ...), but there doesnt seem to be one used in the project?
```
hikari-no-yume: Maybe you can just use a `String` or something? I don't think there's going to be any interesting context, we'll either find the bundle or we won't.

src/fs/bundle.rs
```diff
@@ -0,0 +1,235 @@
...
+    fn find_bundle_path_in_archive(
+        zip: &mut ZipArchive<std::fs::File>,
+    ) -> Result<String, ZipError> {
+        static BUNDLE_NAME_REGEX: Lazy<Regex> =
+            Lazy::new(|| Regex::new(r"Payload/([^/]+)\.app/").unwrap());
```
hikari-no-yume: It's not a big deal, but I think we could match this pattern without needing a regex library.

src/fs/bundle.rs
```diff
@@ -0,0 +1,235 @@
...
+                Ok(buf)
+            }
+        }
+    }
+}
+
+/// Represents a file inside an IPA bundle that can be opened.
+#[derive(Debug)]
+pub struct IpaFileRef {
+    archive: Arc<Mutex<ZipArchive<std::fs::File>>>,
```
hikari-no-yume: Currently this project is _very_ single-threaded, even though it simulates a potentially multi-threaded app, so it'd be fine to use `Rc` instead. I assume we can easily change it back in future? I'm not sure about when it's necessary to use a `Mutex` in Rust, but I assume it has a single-threaded replacement too.

src/fs/bundle.rs
```diff
@@ -0,0 +1,235 @@
...
+        }
+    }
+}
+
+/// Represents an opened file in an IPA bundle.
+pub struct IpaFile {
+    // we need to use a cursor because zip::read::ZipFile doesn't implement Seek
+    // and, generally, seeking in compressed files is hard to achieve
+    // the simplest way to do it is to read the whole file into memory
+    // the target apps should be small enough to fit in memory, right?
```
hikari-no-yume, line 210: Hmm, this might end up a problem with large files (app bundles can contain video files and stuff), but I assume they won't be kept open for too long so it sounds fine for now.

hikari-no-yume (commented):
Hey, thanks a lot! This looks pretty great in general, I appreciate the way you've abstracted things, I'd have done something similar. I didn't look very closely at the new `bundle.rs` file, but I assume it's fine.

I have various comments. If you don't feel like doing all of them, I'm happy to do some of them myself.

src/fs.rs
DCNick3: Yeah, you could not call this constructor with `&str` argument

src/fs.rs
DCNick3: Ah, I knew I missed something! This is why I want a "FileRef" enum variant

.gitignore
DCNick3: It was added by me, because CLion puts some stuff in `.idea` (and if somebody will try to open this in CLion they may accidentally commit some garbage). I can remove it and put it to my local .gitignore if you want though

src/fs.rs
DCNick3: It might be a good idea to have interior mutability (like a `Mutex` or a `Cell`) in fs, so it would always take a `&self`?

.gitignore
hikari-no-yume: I'd prefer that since this might end up being a laundry list of IDEs otherwise, which is a distraction from the main purpose IMO.

src/fs.rs
hikari-no-yume: Huh, not sure how I missed that. I wasn't aware of this detail of Rust, thanks for the fix!

.gitignore
DCNick3: >  laundry list of IDEs

It often is, just look at https://github.com/dotnet/runtime/blob/main/.gitignore

src/fs.rs
hikari-no-yume: Maybe not in this in-line comment section, but if you could elaborate a bit more on why you want it, that'd be great.

src/fs/bundle.rs
DCNick3: Yeah, I think this is it. The zip spec says "files can come in any order", extraction logic in `zip` crate always tries to create parent directories, so I guess this can be possible (even though it's weird)

src/fs/bundle.rs
DCNick3: Hmmm, how are gonna implement threading btw? I tried to design a threading system in my HLE and found it... Non-trivial, so to say

.gitignore
hikari-no-yume: I guess it's a good idea if you have a large project and a lot of novice contributors, but I feel like it's a workaround for people not being familiar with better tools, to be honest. I'd rather keep the `.gitignore` to stuff specific to this project.

src/fs/bundle.rs
DCNick3: Yeah, I figured this too. Just something to be aware of

src/fs/bundle.rs
DCNick3: I don't want to reinvent regex =(

src/fs/bundle.rs
DCNick3: Yeah, based on discussion in #33, I think it's fine to use `String` here

src/fs.rs
hikari-no-yume: I think I would like it if you need `&mut` to “mutate” the FS somehow, but interior mutability would be a nice trick to solve the implementation-sharing problem. We can enforce that invariant within a wrapper function in that case. Changing this should be a separate commit or PR though.

Cargo.lock
DCNick3: Just an empty one won't do (all the compression formats are features :P), but yeah, crypto is probably not something we touchHLE needs

src/fs/bundle.rs
```diff
@@ -0,0 +1,235 @@
+use crate::fs::{FsNode, GuestPath};
```
hikari-no-yume, line 7: Oh right, can you add the MPL-2.0 license header here (please copy it from another file), and a doc comment (a one-liner is fine) summarising what this file is for? The license header is necessary because MPL-2.0 is a per-file rather than whole-project license, so it needs to be explicit which files it applies to, and the doc comments are just a thing I like to have to make the project nice to navigate in `cargo doc`.

src/fs.rs
DCNick3: Sure

Cargo.lock
hikari-no-yume: Oh haha, I assume it's just DEFLATE we need then? (I haven't actually checked…)

DCNick3: Deflate is definitely the most popular zip compression method. I can imagine some files using a different compression, but.. Compile times require sacrifices or smth

src/fs/bundle.rs
hikari-no-yume: That's… fair. I just think Rust usually offers enough nice ways to do things (e.g. the `str` methods and `Iter`) that regex is a bit overkill, and I'm a bit wary of it being deceptively simple sometimes (e.g. unclear greedy vs non-greedy matching behaviour).

I think something like this would work?

```rust
if let Some(name) = path
    .strip_prefix("Payload/")
    .and_then(|path| path.split_once('/'))
    .and_then(|(name, _)| name.strip_suffix(".app")) {
```

src/fs/bundle.rs
hikari-no-yume: There's already a very primitive threading system since Super Monkey Ball needs it. I can explain a bit about how it works in some other channel (I don't want to get too off-topic here), but you can see the guts of it in `main.rs`. Incidentally, I really should move it out of that file, it's getting too complicated.

hikari-no-yume:
Oh right, the README and maybe the help/usage text should be updated after this. I'm happy to do that myself, I like writing documentation.

DCNick3:
I think all the comments are addressed now

DCNick3:
Regarding the `FileRef`: When I first added the `IpaBundleFile` enum variant, I completely forgot to update the places where matches are made over it (and the `is_file` even made it into git).

What I think could have helped is to add another level of indirection (as always): have a "this is a file" enum variant in the `FsNode`  and an inner `FileRef` enum specifying which kind. 

This way the `is_file` problem would not have happened (one enum variant for all files, no need to update)

In the places you actually need to look inside the file (like open), you would first match over the `FsNode` to check whether it's a file and then you would match __exhaustively__ over `FileRef` (as you no longer need "it's a directory" error) and, therefore, changing the `FileRef` enum will give you some compiler errors to tell you what needs updating.

But I don't think it's a good idea to it in this PR

src/fs/bundle.rs
hikari-no-yume: Can you remove the comment?

hikari-no-yume: Wait I'm not sure if GitHub is showing me the latest changes, uhh…

hikari-no-yume:
Yeah, that justification makes sense. I'll think about doing something like that if the enum gets any more complicated than it already is.

src/fs/bundle.rs
DCNick3: It's "outdated". You need to find this place in the changes tab again ig. I think there is also a way to view changes from the last review.

image

src/fs.rs
```diff
@@ -20,15 +20,25 @@
 //! Directories only need a corresponding directory in the host filesystem if
```
hikari-no-yume, line 20: Oh, the paragraph above this is outdated now that we have ZIP files. Uhh, I think I'll try to rewrite it myself later if that's okay.

DCNick3: Sure

src/main.rs
```diff
@@ -282,10 +282,12 @@ impl Environment {
...
         let startup_time = std::time::Instant::now();
 
-        let (bundle, fs) = match bundle::Bundle::new_bundle_and_fs_from_host_path(bundle_path) {
+        let bundle_data = fs::BundleData::open_any(&bundle_path)
+            .map_err(|e| format!("Could not open app bundle: {e}"))?;
+        let (bundle, fs) = match bundle::Bundle::new_bundle_and_fs_from_host_path(bundle_data) {
             Ok(bundle) => bundle,
             Err(err) => {
-                return Err(format!("Application bundle error: {}. Check that the path is to a .app directory. If this is a .ipa file, you need to extract it as a ZIP file to get the .app directory.", err));
+                return Err(format!("Application bundle error: {err}. Check that the path is to a .app directory or an .ipa file."));
```
hikari-no-yume: Super nitpick, I know, but: this usage of “a” versus “an” isn't internally consistent with the message :p

hikari-no-yume:
Compiling this on my machine right now, I'll tell you once I've tested it a bit. I'll also have a look at the handful of IPA files I have to see what compression methods they have.

hikari-no-yume:
`cargo clippy` has three suggestions for me now, maybe you'll want to address them. Otherwise I'll do it myself later.

src/dyld.rs
```diff
@@ -123,7 +123,7 @@ pub fn search_lists<T>(
         .iter()
         .flat_map(|&n| n)
         .find(|&(sym, _)| *sym == symbol)
-        .map(|&(_, ref f)| f)
+        .map(|(_, f)| f)
```
hikari-no-yume: Sorry, super nitpick, can you revert the change in this file? Just because it's nothing to do with your changes.

hikari-no-yume:
I guess I have a newer Rust toolchain than you because there's one more lint for me:

```
warning: use of `ok_or` followed by a function call
  --> src/bundle.rs:36:14
   |
36 |             .ok_or("plist root value is not a dictionary".to_string())?;
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `ok_or_else(|| "plist root value is not a dictionary".to_string())`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call
   = note: `#[warn(clippy::or_fun_call)]` on by default

warning: `touchHLE` (bin "touchHLE") generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 10.52s
```

src/dyld.rs
DCNick3: Whoops :sweat_smile:, just run `cargo clippy --fix`

DCNick3:
I've pushed the fixes for the lints

Just a note: in rust 1.66 a new [uninlined_format_args](https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args) lint was added, which is triggered in... almost the entire project :P (that's why I was kinda ignoring clippy here)

It's machine fixable, so probably want to update the toolchain & run `cargo clippy --fix`

DCNick3:
> I guess I have a newer Rust toolchain than you because there's one more lint for me:

I am on latest stable, 1.67. Still, will fix it

hikari-no-yume:
Ohh wow, I've stubbornly avoided using inlined format args, but maybe it's time for me to switch. Also, ah, for some reason I've never used `cargo clippy --fix`, I guess I've been wasting time :p

DCNick3:
> I've stubbornly avoided using inlined format args

Same, lol. Now that clippy is complaining I have no choice ig

hikari-no-yume:
macOS has an `unzip` utility, I assume the same or a similar tool is available on common Linux distros. I was able to write this one-liner to check a bunch of IPA files' compression methods:

```sh
for i in *.ipa; do unzip -v "$i" | cut -f 3,1 -w | grep -v 'Defl' | grep -v 'Stored' ; done
```

I ran this on all the IPAs I have on my drive: all the ones I redownloaded from the iTunes Store, and a number of others.

As expected, everything is either DEFLATE or uncompressed (aka “Stored”). I think the latter is maybe just for empty files?

So not supporting other compression methods seems pretty safe. 👍

hikari-no-yume:
Oops, I got distracted. Works great in my testing. I was a bit worried it'd be slower than loading a pre-extracted file, but I can't notice the difference in release builds.

I think I'm ready to merge this. My preference is a squash-and-rebase merge type. Do you want to squash all the commits into one yourself (if so, please use `--reset-author` so the author date is more accurate) or shall I do it?

Thanks a lot for this contribution!

DCNick3:
>I was a bit worried it'd be slower than loading a pre-extracted file, but I can't notice the difference in release builds.

The whole game is like 40 megabytes, really small for modern hardware

>Do you want to squash all the commits into one yourself

Sure

hikari-no-yume:
Yeah I guess, but I know from one of my emulator developer friends that even small titles can chug if you have to decompress too often! Luckily Monkey Ball has loading screens.

DCNick3:
(Did I do it right? never squashed this way before...)

hikari-no-yume:
Yeah, looks good to me! My preference is to remove the info in the commit message about which commits were squashed, since that's usually useless extra info in future when reading the commit history, so I'll remove that bit when I cherry-pick the commit, if that's okay.

hikari-no-yume:
Merged as 32d12cc697f8fc62770962b8d5a1fee44dfba677, thanks!

I added a changelog and credited you via your GitHub username in c407ef831b9f67ef0ced294dc30208108b16cf2c. If you'd prefer a different style of credit, please let me know.

PR #41, opened 2023-02-05
Add stubs for various APIs

nitinseshadri:
This PR includes some changes I made in an attempt to get some other apps to run. I was unsuccessful (there are a large number of POSIX functions that are missing), but I hope that this can serve as a helpful starting point for improved compatibility.


.gitignore
```diff
@@ -1,3 +1,4 @@
+.DS_Store
```
hikari-no-yume, line 1: Hi, I recommend adding a git config that globally ignores this file, rather than doing it per-project.

There's a config item in git called `core.excludesfile` that can be used for this.

src/frameworks/openal.rs
```diff
@@ -291,6 +291,15 @@ fn alBufferDataStatic(
     alBufferData(env, buffer, format, data, size, samplerate);
 }
 
+// Apple-specific extension to OpenAL
+fn alcMacOSXMixerOutputRate(_env: &mut Environment, value: ALdouble) {
```
hikari-no-yume, line 285: Ooh. For Super Monkey Ball I was able to get away with returning NULL when it tried to look up the address of this function, but I guess some apps try to call it directly. This is an Apple-specific extension that OpenAL Soft doesn't have, but I assume it's unimportant, so we can probably just make it do nothing, and remove the Super Monkey Ball hack.

hikari-no-yume:
Cool! I think some of these are a bit too stub-y, but there's some interesting stuff in here… I'll try to remember to look at it more later.

hikari-no-yume:
Oh yeah, just curious, what apps were you trying?

nitinseshadri:
Thanks! The apps I tried were:

* Doom for iPhone 1.0 (open-sourced by the developer: https://github.com/id-Software/DOOM-iOS/tree/0cdbbdf96e4f331676b8ce96e20ac4943d0ff0f7)
    * Still crashes due to multiple POSIX functions not being implemented. I stubbed a few more of these (not included in this PR because they're incomplete) but gave up because there were too many.

* Rat on a Skateboard (iOS 2.2.1+)
    * This app tries to initialize a CADisplayLink (among other things) and fails. CADisplayLink was not introduced as public API until iOS 4.0, but it appears that this app got around that on earlier versions by initializing it through NSClassFromString. This is also why I had to implement UIDevice, as the app compares the system version to see if it needs to use NSClassFromString or if it can use the public API.

.gitignore
nitinseshadri: I've been using git for years and still didn't know about this. Thanks!

hikari-no-yume:
I think Wolfenstein RPG uses POSIX stuff too, I'm guessing it was developed by the same team or shares some code.

src/frameworks/openal.rs
nitinseshadri: Yeah. This appears to be used to set the output sample rate. There's a book (from 2010!) which goes into more details about it here:

https://www.google.com/books/edition/Beginning_iPhone_Games_Development/xtrEvAdSOWEC?hl=en&gbpv=1&dq=alcMacOSXMixerOutputRate&pg=PA388&printsec=frontcover

hikari-no-yume:
I apologise if you can't comment etc due to the current interaction pause btw. Feel free to contact me via another means if you want, but I'll probably not respond very quickly.

hikari-no-yume:
I can't find a way to contact you outside of GitHub, can you join this? https://discord.gg/6t7WxQXX

src/frameworks/foundation/ns_string.rs
```diff
@@ -591,7 +591,7 @@ pub fn handle_constant_string(mem: &mut Mem, objc: &mut ObjC, constant_str: id)
         bytes,
         length,
     } = mem.read(constant_str.cast());
-    assert!(flags == 0x7C8); // no idea what this means
+    assert!(flags == 0x7C8); // see: https://lists.llvm.org/pipermail/cfe-dev/2008-August/002518.html
```
hikari-no-yume, line 609: Aha, I suspected it was some CFType thing but knowing the details is great, thanks!

hikari-no-yume:
Would you mind if I just cherry-pick the commits I'm happy with straight to trunk, and then I review the ones I think could be improved?

src/frameworks/openal.rs
hikari-no-yume: Sadly I can't see that, maybe because of my region. I guess it's nothing important though.

nitinseshadri:
Sure, go ahead!

hikari-no-yume:
Can you certify the following? I'm thinking of requiring it for future contributions:

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

src/frameworks/uikit/ui_device.rs
```diff
@@ -14,9 +16,49 @@ pub const UIDeviceOrientationPortrait: UIDeviceOrientation = 1;
...
+          &mut env.mem
+      );
+      env.framework_state.uikit.ui_device.current_device = Some(new);
+      new
+    }
+  }
+
+  - (id) systemVersion { // NSString
+    let system_version = "2.0".to_string();
+    let new = ns_string::from_rust_string(env, system_version);
```
hikari-no-yume: You probably want to use `ns_string::get_static_str` to avoid a memory leak here. I guess the framework state here was an attempt at caching but it looks broken to me.

src/frameworks/uikit/ui_device.rs
```diff
@@ -14,9 +16,49 @@ pub const UIDeviceOrientationPortrait: UIDeviceOrientation = 1;
...
+          this,
+          Box::new(TrivialHostObject),
+          &mut env.mem
+      );
+      env.framework_state.uikit.ui_device.current_device = Some(new);
+      new
+    }
+  }
+
+  - (id) systemVersion { // NSString
```
hikari-no-yume: Style nitpicks:

* I tend to not have a space between the `)` and the method name, but of course it's not important.

* Hmm, I think I'm only using comments to mark argument types rather than return types right now, but maybe I should change that. Perhaps return types could be on the preceding line so that they'd be clear in a case like:

```
// NSString
- (id)methodReturningStringWithString:(id)str { // NSString
```

src/frameworks/foundation/ns_thread.rs
```diff
@@ -13,6 +13,11 @@ pub const CLASSES: ClassExports = objc_classes! {
 
 @implementation NSThread: NSObject
 
++ (f64)threadPriority {
+    log!("TODO: [NSThread threadPriority] (not implemented yet)");
+    return 1.0 // always return 1.0
```
hikari-no-yume: Nitpick: in Rust, the final expression in a block (if not followed by a semicolon) is the return value, so the `return` is redundant and generally not used in this case. The comment is also a bit redundant.

src/objc/properties.rs
```diff
@@ -62,3 +63,32 @@ pub(super) fn objc_setProperty(
...
+//       says that hasStrong is unused, but other versions of the source seem to use it. Not sure what to do here.
+pub(super) fn objc_copyStruct(
+    env: &mut Environment,
+    dest: MutVoidPtr,
+    src: ConstVoidPtr,
+    size: GuestUSize,
+    _atomic: bool,
+    _hasStrong: bool,
+) {
+    // TODO: implement atomic and hasStrong support
```
hikari-no-yume: The source code you reference suggests `_hasStrong` is always unused, so maybe we can ignore it?

As for the atomic, I think it would be good to `assert!(!atomic);` so we'll catch it if an app needs it.

src/objc/properties.rs
```diff
@@ -62,3 +63,32 @@ pub(super) fn objc_setProperty(
...
+    // Right now this is a clone of memmove.
```
hikari-no-yume, line 79: You could actually just call `crate::libc::string::memmove` if you want to.

src/frameworks/foundation/ns_set.rs
```diff
@@ -83,6 +84,12 @@ pub const CLASSES: ClassExports = objc_classes! {
 // TODO: more init methods, etc
 
 // TODO: accessors
+- (id)allObjects {
+    // TODO: actually implement this
+    let object = env.objc.borrow_mut::<SetHostObject>(this).dict.iter_keys().next().unwrap();
```
hikari-no-yume: This looks very close to an actual implementation. I think you would just need to use `.collect()` rather than `.next().unwrap()`.

It wouldn't work with foreign subclasses of NSSet but that's fine I guess.

src/frameworks/foundation/ns_set.rs
```diff
@@ -84,6 +84,12 @@ pub const CLASSES: ClassExports = objc_classes! {
 // TODO: more init methods, etc
 
 // TODO: accessors
+- (id)anyObject {
+    // TODO: actually implement this
+    let object = env.objc.borrow_mut::<SetHostObject>(this).dict.iter_keys().next().unwrap();
```
hikari-no-yume: This also looks very close to an actual implementation. If you `match`'d the result of `.next()` you could then return `nil` if it returns `None`.

This wouldn't work with foreign subclasses but that's a common issue in our Foundation right now.

src/libc/string.rs
```diff
@@ -111,6 +111,65 @@ fn strcmp(env: &mut Environment, a: ConstPtr<u8>, b: ConstPtr<u8>) -> i32 {
...
+
+    let mut offset = 0;
+    loop {
+        let char_string = env.mem.read(string + offset);
+
+        if char_string == 0 {
+            return ConstPtr::from_bits(0); // NULL
+        }
+
+        match char_string.cmp(&char_substring) {
```
hikari-no-yume: You could simplify this to just do `if char_string == char_substring {`.

src/frameworks/opengles/gles1_on_gl2.rs
```diff
@@ -665,6 +665,10 @@ impl GLES for GLES1OnGL2 {
         );
         gl21::TexParameteri(target, pname, param);
     }
+    unsafe fn TexParameterf(&mut self, target: GLenum, pname: GLenum, param: GLfloat) {
+        assert!(target == gl21::TEXTURE_2D);
```
hikari-no-yume, line 669: Huh, what parameter is the app trying to set? Looking at https://registry.khronos.org/OpenGL-Refpages/es1.1/xhtml/ I think all the parameters are integers, so maybe it's relying on the OpenGL implementation doing a float-to-integer conversion. If so, you can copy the assert from `TexParameteri`, but add a comment that they are all integers.

src/libc/pthread/thread.rs
```diff
@@ -173,11 +174,18 @@ fn pthread_mach_thread_np(env: &mut Environment, thread: pthread_t) -> mach_port
     host_object.thread_id.try_into().unwrap()
 }
 
+// POSIX usleep
+fn usleep(_env: &mut Environment, useconds: u32) {
+    let duration = Duration::from_millis(u64::from(useconds) / 1000);
```
hikari-no-yume: Rust offers a `Duration::from_micros` method so there's no need for a conversion here.

src/libc/pthread/thread.rs
```diff
@@ -173,11 +174,18 @@ fn pthread_mach_thread_np(env: &mut Environment, thread: pthread_t) -> mach_port
...
+    std::thread::sleep(duration);
```
hikari-no-yume, line 180: I'm probably not going to merge this because it's, unfortunately, not suited to the threading system touchHLE uses right now. Instead of real (host OS) threads, all the threads are handled by a single thread in a co-operative fashion. In order to properly “sleep”, I'd want to do it in a way where the other threads can keep running. I guess I'd need to add a method to the `Environment` type for doing that, where it'll run the other threads for a while and then return to the original thread once the duration is over.

By the way, do you know what duration the app you were testing was sleeping for? I'm curious whether it's doing a short or a long sleep.

src/frameworks/foundation/ns_string.rs
```diff
@@ -435,6 +435,18 @@ pub const CLASSES: ClassExports = objc_classes! {
     new_string
 }
 
+- (id)lastPathComponent { // NSString*
+    // TODO: avoid copying
```
hikari-no-yume: I don't think we can actually avoid copying in this case, so you can remove this comment.

src/frameworks/foundation/ns_string.rs
```diff
@@ -435,6 +435,18 @@ pub const CLASSES: ClassExports = objc_classes! {
...
+    let self_as_rust_string = &to_rust_string(env, this);
+    let path = GuestPath::new(self_as_rust_string);
+    let file_name = path.file_name().unwrap_or_else(|| {
+        ""
```
hikari-no-yume: You can use `.unwrap_or("");`

src/frameworks/core_graphics/cg_color_space.rs
```diff
@@ -48,6 +48,19 @@ pub fn CGColorSpaceCreateWithName(env: &mut Environment, name: CFStringRef) -> C
...
 }
 
+pub fn CGColorSpaceCreateDeviceRGB(env: &mut Environment) -> CGColorSpaceRef {
+    let isa = env
+        .objc
+        .get_known_class("_touchHLE_CGColorSpace", &mut env.mem);
+    env.objc.alloc_object(
+        isa,
+        Box::new(CGColorSpaceHostObject {
+            name: kCGColorSpaceGenericRGB,
```
hikari-no-yume, line 59: I was curious what this actually does, so I tested it on macOS:

```objc
#import "Foundation/Foundation.h"
#import "CoreGraphics/CoreGraphics.h"

int main(void)
{
    CGColorSpaceRef cs = CGColorSpaceCreateDeviceRGB();
    NSLog(@"%@", CGColorSpaceCopyName(cs));
}
```

Looks like the result is some undocumented constant:

```
$ ./CGColorSpaceCreateDeviceRGB 
2023-02-10 14:03:28.464 CGColorSpaceCreateDeviceRGB[10366:143295] kCGColorSpaceDeviceRGB
```

I was hoping it might turn out to be an alias for generic RGB depending on the system, but I guess not. Btw, that constant doesn't seem to be in the API, so I don't think it's something we're missing.

Maybe add a comment like `// TODO: figure out what characteristics this actually has on an iPhone`, because I assume we don't know if this is e.g. sRGB versus Linear RGB.

src/frameworks/foundation/ns_object.rs
```diff
@@ -16,10 +16,34 @@
...
+use crate::Environment;
+
+pub(super) fn NSClassFromString(
+    env: &mut Environment,
+    _class_name: id, // NSString
+) -> Class {
+    // TODO: actually implement this
+    println!("Warning: NSClassFromString not implemented, app will crash.");
+    return msg_class![env; NSObject alloc];
+}
```
hikari-no-yume, line 33: I won't merge these as-is because I think they're a bit too hacky and we should just implement them for real. For reference:

* `NSClassFromString` can use `env.objc.get_known_class`. Note that will panic if it doesn't find a class, but that's probably best for the time being.
* `NSLog` can share code with `printf`, though there are some small differences in how format strings work for the two.
* `respondsToSelector:` can actually check if the selector is supported. Some changes in `src/objc/` might be needed. We'll probably want to have some kind of warning if it returns `false` when used on a host class, though, as it's probably the app checking for functionality that we haven't implemented yet.

But to be clear, I don't expect you to implement these unless you want to. I'm sure they're common enough that I'll get round to them eventually.

hikari-no-yume:
Thanks for your contributions, I can see you've put in a lot of work. Sorry for being so picky…

I have a local branch with some cherry-picks from this that I'll merge into trunk at some point:

```
$ git log -6 --pretty=oneline
bc3e81eca6c8671c00d0b6e25fe60fb9fe7985fb (HEAD -> nitinseshadri-cherry-picks) Add explanation for 0x7C8 constant
90fabafb88ad03095396d3f5cd6c421779214680 Remove Monkey Ball-specific alcMacOSXMixerOutputRate hack
3bc75111f1f3e17ec4122be9a775d241251d5c1e Add Apple-specific alcMacOSXMixerOutputRate to OpenAL
475bb5348613f616a71507a66f4d33fff127b3e9 Add support for floats as format arguments in printf et al.
0ba998acee445956dc496ceb35a971e74203b043 Add some audio session functions to AudioToolbox
3183b792f5b4a3f757ee16c56dffa663127a5275 Add support for landscape right orientation
```

In general my policy is to only use stubs where the action or result value of the function/method is not likely to be important. I'm worried that otherwise, the execution flow of the emulated app is going to diverge in weird ways and strange bugs will appear. This policy, and some stylistic things, are why I've cherry-picked so few of these.

hikari-no-yume:
Btw, I have a private Discord server specifically for development discussion, if you are interested. You can join it if you want to (email me for an invite).

src/frameworks/opengles/gles1_on_gl2.rs
```diff
@@ -669,6 +669,9 @@ impl GLES for GLES1OnGL2 {
         assert!(target == gl21::TEXTURE_2D);
         gl21::TexParameterf(target, pname, param);
     }
+    unsafe fn TexEnvi(&mut self, target: GLenum, pname: GLenum, param: GLint) {
+        gl21::TexEnvi(target, pname, param);
```
hikari-no-yume, line 677: Can you add an assert for the `pname` values? Because OpenGL ES 1.1 and OpenGL 2.1 aren't exact supersets or subsets of eachother, I try to check the app doesn't use something unsupported.

What I usually do is go to https://registry.khronos.org/OpenGL-Refpages/es1.1/xhtml/ and https://registry.khronos.org/OpenGL-Refpages/gl2.1/ and look up the documentation for the two versions of the function, then compare them to see what the common subset of enum values is. Then I add an assert that the value is in that common subset.

src/frameworks/foundation/ns_object.rs
hikari-no-yume: Actually, I am probably going to implement `NSLog` today, so don't worry about that one. :)

src/frameworks/foundation/ns_string.rs
hikari-no-yume: Sorry, what am I saying, of course we can avoid copying for the `to_rust_string` part (eventually). Ignore me.

src/frameworks/core_graphics/cg_color_space.rs
nitinseshadri: I added a comment

src/libc/pthread/thread.rs
nitinseshadri: The app that does this is Doom. It seems to be a very short sleep, only 1ms: https://github.com/id-Software/DOOM-iOS/search?q=usleep

src/libc/string.rs
```diff
@@ -148,20 +148,16 @@ fn strstr(env: &mut Environment, string: ConstPtr<u8>, substring: ConstPtr<u8>)
...
-                        return string + offset;
-                    }
-                    if char_string_plusone != char_substring_plusone {
-                        break;
-                    }
+        if char_string == char_substring {
+            let char_substring_plusone = env.mem.read(substring + 1);
+            let char_string_plusone = env.mem.read(string + offset + 1);
+
+            loop {
```
hikari-no-yume, line 163: This doesn't look right, you're not incrementing at all within the loop…

nitinseshadri:
What are your thoughts on the changes related to -[UIView layoutSubviews] and -[UIApplicationDelegate applicationDidBecomeActive:]?

hikari-no-yume:
I actually tried adding `applicationDidBecomeActive:` for Super Monkey Ball, but it didn't affect whether the app ran correctly in any way, and if I remember correctly it caused it to call an extra API that I didn't feel like implementing. So it was just for convenience that I skipped it. Likewise, Super Monkey Ball doesn't seem to need `layoutSubviews`.

But I'm sure we'll have to call those eventually.

I guess adding a default implementation of `layoutSubviews` is harmless.

nitinseshadri:
> I actually tried adding `applicationDidBecomeActive:` for Super Monkey Ball, but it didn't affect whether the app ran correctly in any way, and if I remember correctly it caused it to call an extra API that I didn't feel like implementing. So it was just for convenience that I skipped it.

I think I ended up implementing those. I think it was glDeleteFramebuffersOES() and glDeleteRenderbuffersOES(). The app also calls those in layoutSubviews and in applicationWillTerminate.

src/frameworks/openal.rs
nitinseshadri: Yeah. The book just talks about how setting it to the same sample rate as your audio files can possibly increase performance by avoiding the need for sample rate conversions.

src/objc/properties.rs
nitinseshadri: I wasn't sure if it made sense to call string's memmove. Maybe the memmove function should be moved to the top-level libc module?

src/frameworks/opengles/gles1_on_gl2.rs
nitinseshadri: GL_TEXTURE_MAX_ANISOTROPY_EXT, which is part of the anisotropic filtering ubiquitous extension to OpenGL (in general, not just ES 1.1): https://www.khronos.org/opengl/wiki/Ubiquitous_Extension

Apparently, iPhone OS has supported it since 2.0: https://developer.apple.com/documentation/opengles/gl_texture_max_anisotropy_ext?language=objc

GL_MAX_TEXTURE_MAX_ANISOTROPY_EXT is also part of this, so I would need to add that to the assert as well.

src/frameworks/opengles/gles1_on_gl2.rs
nitinseshadri: The only difference appears to be that ES 1.1 has GL_COORD_REPLACE_OES, while 2.1 has GL_COORD_REPLACE. I can't seem to find any info about the difference between the two, though.

src/frameworks/opengles/gles1_on_gl2.rs
nitinseshadri: So it looks like the OpenGL bindings don't have constants for either of these. What would be the correct way to add them?

src/frameworks/opengles/gles1_on_gl2.rs
nitinseshadri: Interestingly, these parameter names both have the same value (0x8862 or 34914). Could they be considered equivalent?

src/libc/string.rs
nitinseshadri: Whoops! Should be fixed now.

src/frameworks/foundation/ns_object.rs
nitinseshadri: I'm probably going to leave this as-is for now.

Ideally, NSClassFromString should return `nil` if the class does not exist, because some apps use NSClassFromString when checking if an API exists or not. In fact, even Apple did this in early sample code for iPhone OS (see EAGLView.m): https://developer.apple.com/library/archive/samplecode/GLES2Sample/Introduction/Intro.html#//apple_ref/doc/uid/DTS40009188-Intro-DontLinkElementID_2

src/libc/string.rs
```diff
@@ -142,23 +142,33 @@ fn strstr(env: &mut Environment, string: ConstPtr<u8>, substring: ConstPtr<u8>)
...
-            let char_string_plusone = env.mem.read(string + offset + 1);
+            // We found a match - let's compare the rest of the substring to see if it matches too.
+
+            let mut inner_offset = 1;
 
             loop {
+                let char_substring_inner = env.mem.read(substring + inner_offset);
+                let char_string_inner = env.mem.read(string + offset + inner_offset);
+
                 if char_substring == 0 {
```
hikari-no-yume, line 167: Shouldn't this also check `char_substring_inner`?

I think this implementation is overcomplicated by trying to match the first character separately from the rest of the characters. It can probably be simplified to something like:

```rust
let mut offset = 0;
loop {
    let mut inner_offset = 0;
    loop {
        let char_string = env.mem.read(string + offset + inner_offset);
        let char_substring = env.mem.read(substring + inner_offset);
        if char_substring == b'\0' {
            return string + offset;
        } else if char_string == b'\0' {
            return Ptr::null();
        } else if char_string != char_substring {
            break;
        } else {
            offset += 1;
        }
    }
    offset += 1;
}
```

Note that I use `b'\0'` instead of `0` to emphasise it's a character, but that's just personal preference…

Hmm, I may as well commit that as my own code, I guess I'll do that once I've tested it a bit.

nitinseshadri: That makes sense. I was trying to do an implementation similar to the one in the link I reference (https://opensource.apple.com/source/tcl/tcl-10/tcl/compat/strstr.c.auto.html), but this seems to be a simpler one.

src/frameworks/opengles/gles1_on_gl2.rs
hikari-no-yume: Have a look at `src/gl_bindings/build.rs`. It has a list of extensions to include in the generated bindings.

src/frameworks/foundation/ns_object.rs
hikari-no-yume: Yeah, I guess the real implementation of that would need to be full of special cases.

src/frameworks/opengles/gles1_on_gl2.rs
hikari-no-yume: I think so? I am not sure why some things have that suffix in OpenGL ES 1.1 but not in OpenGL 2.1. It might be because OpenGL ES 1.1 is based on OpenGL 1.5, which maybe didn't have some of these enum values.

I should probably add a note about it somewhere, but I assume in many places that enum values match between ES and normal GL.

src/libc/string.rs
hikari-no-yume: If it's derived from the algorithm in public source code, we shouldn't use it anyway due to the copyright risk.

src/frameworks/opengles/gles1_on_gl2.rs
nitinseshadri: Would the assert still be needed?

hikari-no-yume:
I notice you're appending new commits to correct issues in older commits. Perhaps because my old employer used Gerrit, I prefer to modify the old commits instead by editing the git history, so each commit is complete and self-contained. I can recommend `git rebase -i` if you're not familiar with it.

You don't have to do this yourself though, I'm going to do it anyway when merging.

src/frameworks/opengles/gles1_on_gl2.rs
hikari-no-yume: I would still prefer to have the assert. I want to ensure that if the app uses something we don't expect, we catch it right away, rather than it causing e.g. an OpenGL error that will only be caught the next time the app calls `glGetError()`.

hikari-no-yume: Actually, I need to implement part of the `glTexEnv` family now for the app I'm currently trying to get running, so I may as well implement the full thing and save you the effort.

src/frameworks/foundation/ns_set.rs
nitinseshadri: Should be fixed now, I rewrote the git history as well.

src/frameworks/foundation/ns_set.rs
nitinseshadri: Should be fixed now, I rewrote the git history as well

hikari-no-yume:
It's no problem for me, but you may want to rebase against the latest trunk so that you don't get weird duplicates of some of the commits merged in from trunk.

nitinseshadri:
Should be good now. I rewrote the history so it was a bit clearer, but you might need to re-cherry-pick things as a result.

hikari-no-yume:
Thanks a lot, this looks much cleaner!

hikari-no-yume:
Uhh, did you miss my comment about the Developer Certificate of Origin from earlier? I don't need you to actually sign all the commits or anything, but if you can at least tell me you agree with all of it, that's reassuring. (Sorry, I'm a bit paranoid with this particular project…)

nitinseshadri:
Yes, I agree with all parts of the Developer Certificate of Origin.

src/libc/string.rs
```diff
@@ -111,6 +111,71 @@ fn strcmp(env: &mut Environment, a: ConstPtr<u8>, b: ConstPtr<u8>) -> i32 {
     }
 }
 
+fn strncmp(env: &mut Environment, a: ConstPtr<u8>, b: ConstPtr<u8>, n: GuestUSize) -> i32 {
+    let mut offset = 0;
+    loop {
```
hikari-no-yume, line 120: I think you need to add

```rust
if n == 0 {
    return 0;
}
```

before this loop, to ensure no characters get compared in that case.

src/frameworks/foundation/ns_timer.rs
```diff
@@ -118,6 +118,12 @@ pub const CLASSES: ClassExports = objc_classes! {
     env.objc.borrow::<NSTimerHostObject>(this).due_by.is_some()
 }
 
+-(())invalidate {
+    // Remove the timer from the run loop.
+    let run_loop: id = msg_class![env; NSRunLoop currentRunLoop];
+    ns_run_loop::remove_timer(env, run_loop, this);
```
hikari-no-yume, line 125: I think you also need to ensure `isValid` above will return `false` in this case, perhaps by setting `due_by` to `None`.

src/frameworks/foundation/ns_string.rs
```diff
@@ -435,6 +435,15 @@ pub const CLASSES: ClassExports = objc_classes! {
     new_string
 }
 
+- (id)lastPathComponent { // NSString*
+    let self_as_rust_string = &to_rust_string(env, this);
+    let path = GuestPath::new(self_as_rust_string);
+    let file_name = path.file_name().unwrap_or("");
```
hikari-no-yume: I just noticed that the path `"/"` is a special case, I guess that needs to be accounted for.

hikari-no-yume:
I've merged most of these now, thanks a lot! I've commented on the remaining ones that I noticed needed fixes, and the rest are things that I don't want to merge (or possibly things I missed).

src/objc/properties.rs
hikari-no-yume: This reminded me that I already wanted to replace my `memmove` implementation with Rust's built-in one,  so I did that now: bbf6c68d820d75849febd1747c4881006bd9f06f.

That's not a `libc` function so you can use it here, perhaps?

src/libc/string.rs
hikari-no-yume: Did this now: 741c509061ee5ce3291cd18d0b81ddcfa237612b.

nitinseshadri:
> I've merged most of these now, thanks a lot! I've commented on the remaining ones that I noticed needed fixes, and the rest are things that I don't want to merge (or possibly things I missed).

I think I addressed all the comments; the only things which I think you missed were
* Add stub for [UIView setExclusiveTouch:] (unless you don't want to merge this)
* Add [UIView layoutSubviews:] and [UIApplicationDelegate applicationDidBecomeActive:] (works with Super Monkey Ball now that glDeleteFramebuffersOES and glDeleteRenderbuffersOES are implemented, unless there is another reason to not merge this. Some other apps appear to use both places to do initialization so I think it would be worth merging.)

src/objc/properties.rs
nitinseshadri: Yep. I'm having trouble rebasing onto latest trunk, though, so would you be able to make that change?

hikari-no-yume:
> Add stub for [UIView setExclusiveTouch:] (unless you don't want to merge this)

I want to skip this for the time being since I'm pretty sure we wouldn't use the right behaviour. It might be trivial to do a real (if hacky) implementation though by storing which view has exclusive touch somewhere?

> Add [UIView layoutSubviews:] and [UIApplicationDelegate applicationDidBecomeActive:] (works with Super Monkey Ball now that glDeleteFramebuffersOES and glDeleteRenderbuffersOES are implemented, unless there is another reason to not merge this. Some other apps appear to use both places to do initialization so I think it would be worth merging.)

Oh, sorry, I didn't realise you'd re-enabled those. Hmm, sure, I'll probably merge those soon. We should probably also send `applicationWillResignActive:` when quitting.

src/frameworks/foundation/ns_string.rs
```diff
@@ -435,6 +435,21 @@ pub const CLASSES: ClassExports = objc_classes! {
...
+        autorelease(env, new_string);
+        new_string
+    } else {
+        let path = GuestPath::new(self_as_rust_string);
+        let file_name = path.file_name().unwrap_or("");
+        let new_string = from_rust_string(env, String::from(file_name));
+        autorelease(env, new_string);
+        new_string
+    }
+}
```
hikari-no-yume, line 451: Ah, sorry, the more I look at this, the more I notice it's not correct, and I'm not entirely sure how to fix it… are you okay if I just don't merge it for now?

nitinseshadri: That's fine. In what ways is it incorrect?

hikari-no-yume: Just staring at https://developer.apple.com/documentation/foundation/nsstring/1416528-lastpathcomponent?language=objc, I think it's always incorrect to return `""` unless the original string is `""`? I also realised the implementation in `fs.rs` is probably wrong in edge cases like `/foo/` too.

It probably works okay most of the time but eventually someone (probably me) needs to write unit tests for all the path component edge cases in `fs.rs` and make sure they all match Apple's behaviour.

nitinseshadri:
> Oh, sorry, I didn't realise you'd re-enabled those. Hmm, sure, I'll probably merge those soon. We should probably also send `applicationWillResignActive:` when quitting.

Thanks! `applicationWillResignActive:` isn't sent when quitting, it's only sent when there is a temporary interruption (e.g. phone call, text message), so I don't think we'd need to send it.

hikari-no-yume:
Oh really? Hmm, maybe I need to re-read the application lifecycle docs.

I've now cherry-picked what I think are all the remaining ones:

```
$ git log -4 --pretty=oneline
a295b05cf7ee1a44350acb49b8f273b53dded2ec (HEAD -> trunk) Add layoutSubviews and applicationDidBecomeActive:
7174fc62b82b408d48c0f81fa52dfd4cb1a22da2 Add basic implementation of objc_copyStruct
51912f208e25c4d13ac55003bc9cc1d359864d7a Add libc strncmp
d8a6ea289b12e6fff65bf8ecc538c794b4c89ef3 Add [NSTimer invalidate]
```

The only one I didn't cherry-pick is the path component thing.

src/frameworks/foundation/ns_string.rs
nitinseshadri: Yeah, you're absolutely right. I wrote some quick Swift code in a playground to test this with the example cases in the Rust docs[1], and NSString does things very differently. So this is something that will need to be revisited later.

```swift
import Foundation

let string: NSString = "foo.txt/.//"
// note: explicitly initialize as NSString instead of Swift String type

print("last path component: \(string.lastPathComponent)")

// This returns ".", while the corresponding Rust implementation will return "foo.txt".
```

[1]https://doc.rust-lang.org/std/path/struct.Path.html#method.file_name

hikari-no-yume: Oh no, I didn't even consider the problem of a mismatch between Rust's `Path` and `NSString`/`NSURL`! Thanks for pointing that out.

hikari-no-yume:
I've acknowledged your changes in the changelog and credited you via your GitHub username: f061f63e6e9f4e6864d4e4a2b31352984024afb7.

If you'd prefer to be credited in a different way, please let me know!

nitinseshadri:
Thank you!

upintheairsheep:
> 

I personally believe the discord server should be open to the public, but read only unless you are an actual developer. That way enthusiasts can check in with development, and you will not have to deal with spam.

hikari-no-yume:
@upintheairsheep My time isn't free and I don't want to have to deal with tons of questions from novice users all the time. With that said, I understand what you're asking for. I would like it to be easy to keep up with new releases and so on. I'm just not sure what way to do it.

src/frameworks/foundation/ns_string.rs
hikari-no-yume: A little update, in case it's interesting. The app I'm currently interested in also needed this method, so I've now done my own implementation, which I think is correct: 09c749c5f514a1cc73bb2ca296e14a4653d7f5a5

PR #49, opened 2023-02-10
Free apps also count as piracy, even though most do not think that

upintheairsheep:


hikari-no-yume:
Thanks for the reminder. I decided to go with a different wording but I credited you for reminding me to do it.

PR #51, opened 2023-02-11
Use LC_UNIXTHREAD to locate the entry point

RealSupremium:
Seems like old apps do consistently name the entry point `Start`, but for newer apps (around iOS 5 and up), it's different.

Does this change really matter? Probably not. Does it hurt? I don't think so.

### DIFF
--- a/src/mach_o.rs
+++ b/src/mach_o.rs
@@ -20,7 +20,7 @@
 
 use crate::fs::{Fs, GuestPath};
 use crate::mem::{Mem, Ptr};
-use mach_object::{DyLib, LoadCommand, MachCommand, OFile, Symbol, SymbolIter};
+use mach_object::{DyLib, LoadCommand, MachCommand, OFile, Symbol, SymbolIter, ThreadState};
 use std::collections::HashMap;
 use std::io::{Cursor, Seek, SeekFrom};
 
@@ -38,6 +38,8 @@ pub struct MachO {
     /// List of addresses and names of external relocations for the dynamic
     /// linker to resolve.
     pub external_relocations: Vec<(u32, String)>,
+    /// Thread state to initialize with (LC_UNIXTHREAD)
+    pub thread_state: ThreadState,
 }
 
 #[derive(Debug)]
@@ -212,6 +214,7 @@ impl MachO {
         let mut exported_symbols = HashMap::new();
         let mut indirect_undef_symbols: Vec<Option<String>> = Vec::new();
         let mut external_relocations: Vec<(u32, String)> = Vec::new();
+        let mut thread_state: Option<ThreadState> = None;
 
         for MachCommand(command, _size) in commands {
             match command {
@@ -372,6 +375,9 @@ impl MachO {
                 LoadCommand::DyldInfo { .. } => {
                     log!("Warning! DyldInfo is not handled.");
                 }
+                LoadCommand::UnixThread { state, .. } => {
+                    thread_state = Some(state);
+                }
                 _ => (),
             }
         }
@@ -420,6 +426,7 @@ impl MachO {
             sections,
             exported_symbols,
             external_relocations,
+            thread_state: thread_state.unwrap(),
         })
     }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -340,84 +340,86 @@ impl Environment {
             };
         }
 
-        let entry_point_addr = *executable.exported_symbols.get("start").ok_or_else(|| {
-            "Mach-O file has no 'start' symbol, perhaps it is not an executable?".to_string()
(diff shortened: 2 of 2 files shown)


hikari-no-yume:
Thanks for figuring this out!

src/main.rs
```diff
@@ -340,84 +340,86 @@ impl Environment {
             };
         }
 
-        let entry_point_addr = *executable.exported_symbols.get("start").ok_or_else(|| {
-            "Mach-O file has no 'start' symbol, perhaps it is not an executable?".to_string()
-        })?;
-        let entry_point_addr = abi::GuestFunction::from_addr_with_thumb_bit(entry_point_addr);
+        if let mach_object::ThreadState::Arm { __pc, .. } = executable.thread_state {
```
hikari-no-yume, line 343: Can you use `let … else {` so we don't need to indent the rest of the function? (Though see my comment in the other file, that would make this redundant.)

src/mach_o.rs
```diff
@@ -38,6 +38,8 @@ pub struct MachO {
     /// List of addresses and names of external relocations for the dynamic
     /// linker to resolve.
     pub external_relocations: Vec<(u32, String)>,
+    /// Thread state to initialize with (LC_UNIXTHREAD)
+    pub thread_state: ThreadState,
```
hikari-no-yume, line 42: If possible I'd prefer not to expose stuff from the mach_object crate outside of this module. Can we extract the bit we are interested in (the entry point address) inside this file instead?

src/mach_o.rs
```diff
@@ -372,6 +375,9 @@ impl MachO {
                 LoadCommand::DyldInfo { .. } => {
                     log!("Warning! DyldInfo is not handled.");
                 }
+                LoadCommand::UnixThread { state, .. } => {
```
hikari-no-yume, line 378: Hmm, it looks like there's also `LoadCommand::EntryPoint`, but I haven't checked if iPhone OS/iOS apps use this or when they started doing so.

hikari-no-yume: Another thing I'm wondering about is whether there can be several of these load commands? Does the app always have exactly one initial thread? I think the safest thing would be to do something like:

```rust
// Assumption: there is only one initial thread
assert!(thread_state.is_none());
thread_state = Some(state);
```

src/main.rs
hikari-no-yume: We should probably also check the other values in the initial state conform to expectations, in case the app is actually relying on them.

hikari-no-yume:
FWIW I'd be happy to take the idea from this pull request and implement this myself in the way I suggest, if you'd prefer that. I'd credit you for the research though. :)

src/main.rs
RealSupremium: The other values that ThreadState has are seemingly important.
image
But I have not seen those used by an app, as they're all zeros except __pc.

src/mach_o.rs
RealSupremium: That is a good idea. Though, I mentioned in an earlier comment that ThreadState may potentially have some additional important stuff for starting execution. For now, I would be happy with just the address being exposed. Ideally, in the future we would want to employ those extra parameters.

src/main.rs
hikari-no-yume: In that case we can just have an assert something like this:

```
assert!(&_r == &[0; 13] && __sp == 0 && __lr == 0 && __cpsr == 0);
```

src/mach_o.rs
RealSupremium: `LoadCommand::EntryPoint` is the new version of `LoadCommand::UnixThread`. Seems like Apple switched to it at the same time as they went 64-bit.

I believe it's only supposed to be one. (https://opensource.apple.com/source/xnu/xnu-2050.18.24/EXTERNAL_HEADERS/mach-o/loader.h)
> For executable objects that are unix processes there will be one thread_command (cmd == LC_UNIXTHREAD) created for it by the link-editor. This is the same as a LC_THREAD, except that a stack is automatically created (based on the shell's limit for the stack size).  Command arguments and environment variables are copied onto that stack.

hikari-no-yume: Ah, if it's a 64-bit thing then I guess we don't have to worry about it.

RealSupremium:
Yeah, I would rather you implement this idea. I have practically no experience with Rust. I am only recently convinced to work in Rust now thanks to you.

I appreciate the thorough review.

hikari-no-yume:
Alright, thanks again for figuring this out for me. :)

I'm happy if this project gets more people to try out Rust, it's proved to be a very nice language for a project like this.

PR #63, opened 2023-02-17
Allow launching apps that do initialization in code, fix "Could not read icon file" issue, and implement exit()

nitinseshadri:
Some apps don't have main nib files and instead choose to do their view initialization and layout in code (probably because nibs, xibs, and storyboards don't really play nice with source control). This additionally means that they have to provide their app delegate class name to UIApplicationMain(). This PR adds basic support for those apps, supporting only scenarios like this one:

```objc
- (void)applicationDidFinishLaunching:(UIApplication *)application {
    window = [[UIWindow alloc] initWithFrame:[[UIScreen mainScreen] bounds]];
    glView = [[EAGLView alloc] initWithFrame:[[UIScreen mainScreen] bounds]];
    [window addSubview:glView];
    [window makeKeyAndVisible];
    [glView startAnimation];
}
```

In addition, this PR fixes some cases of the "Could not read icon file" issue. Some apps don't include the extension of their icon file in their Info.plist, so we add ".png" to the end if there is no extension provided. A logical extension of this would be to look for images suffixed with `@2x` (e.g. `Icon@2x.png`) once Retina mode is supported.

Finally, this PR implements exit().

### DIFF
--- a/src/frameworks/uikit/ui_application.rs
+++ b/src/frameworks/uikit/ui_application.rs
@@ -128,7 +128,8 @@ pub(super) fn UIApplicationMain(
     principal_class_name: id, // NSString*
     delegate_class_name: id,  // NSString*
 ) {
-    if principal_class_name != nil || delegate_class_name != nil {
+    if principal_class_name != nil {
+        // Generally, applications should not subclass UIApplication.
         unimplemented!()
     }
 
@@ -146,7 +147,20 @@ pub(super) fn UIApplicationMain(
         // The delegate must have been created by this point.
         // While notionally UIApplication does not retain its delegate (see
         // `setDelegate:` above), we do have to retain this first one.
-        let delegate: id = msg![env; ui_application delegate];
+        let delegate: id = {
+            if delegate_class_name == nil {
+                msg![env; ui_application delegate]
+            } else {
+                let delegate_class_name_as_rust_string =
+                    ns_string::to_rust_string(env, delegate_class_name);
+                let delegate_class = env
+                    .objc
+                    .get_known_class(&delegate_class_name_as_rust_string, &mut env.mem);
+                let delegate_object = msg![env; delegate_class new];
+                () = msg![env; ui_application setDelegate:delegate_object];
+                delegate_object
+            }
+        };
         assert!(delegate != nil); // should have been set by now
         retain(env, delegate);
 
--- a/src/frameworks/uikit/ui_nib.rs
+++ b/src/frameworks/uikit/ui_nib.rs
@@ -170,8 +170,15 @@ pub const CLASSES: ClassExports = objc_classes! {
 /// return [nib instantiateWithOwner:[UIApplication sharedApplication]
 ///                     optionsOrNil:nil];
 /// ```
-pub fn load_main_nib_file(env: &mut Environment, _ui_application: id) {
-    let path = env.bundle.main_nib_file_path();
+/// Returns true if the main nib file was loaded, false if not.
+pub fn load_main_nib_file(env: &mut Environment, _ui_application: id) -> bool {
+    let path = match env.bundle.main_nib_file_path() {
+        Some(path) => path,
+        None => {
+            // The application doesn't have a main nib file.
+            return false;
+        }
+    };
 
     let data = env.fs.read(path).unwrap();
 
@@ -198,4 +205,6 @@ pub fn load_main_nib_file(env: &mut Environment, _ui_application: id) {
     }
 
     release(env, unarchiver);
+
+    return true;
 }
--- a/src/frameworks/uikit/ui_view.rs
+++ b/src/frameworks/uikit/ui_view.rs
@@ -67,7 +67,26 @@ pub const CLASSES: ClassExports = objc_classes! {
     env.objc.get_known_class("CALayer", &mut env.mem)
 }
 
-// TODO: initWithFrame:, accessors, etc
+- (id)initWithFrame:(CGRect)frame { // CGRect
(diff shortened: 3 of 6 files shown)


nitinseshadri:
Looks like `exit()` has been implemented already. Would be nice to add the following code comment to it though:
```rust
// Apple explicitly says to not call the exit() function. Some apps do it anyway.
// See https://developer.apple.com/library/archive/qa/qa1561/_index.html
```

I'm also printing out the exit code in my implementation:
```rust
println!("Application exited with code {}.", code);
```

hikari-no-yume:
I actually just changed the icon loading code to tolerate the case where it can't find the icon, so I guess you'll have to rebase on top of that. Thanks for figuring out what was wrong there.

I'll have a look at this later. I have some similar local changes for handling apps without nib files (didn't polish them because they weren't enough to get the app I was interested in to work), I'll have to compare them to your changes.

hikari-no-yume:
Ah, sorry, I see your icon changes don't conflict with mine.

src/frameworks/uikit/ui_application.rs
```diff
@@ -146,7 +147,20 @@ pub(super) fn UIApplicationMain(
...
+                let delegate_class_name_as_rust_string =
+                    ns_string::to_rust_string(env, delegate_class_name);
+                let delegate_class = env
+                    .objc
+                    .get_known_class(&delegate_class_name_as_rust_string, &mut env.mem);
+                let delegate_object = msg![env; delegate_class new];
+                () = msg![env; ui_application setDelegate:delegate_object];
+                delegate_object
+            }
+        };
```
hikari-no-yume, line 163: Oh, hmm, when I implemented this I constructed the delegate before loading the nib, but here you do it after.

It's unclear to me whether you can have a named delegate class _and_ a nib file. If you can, I think we might need to construct the delegate first, in case the nib file tries to create connections to it? I'm completely guessing here though.

src/frameworks/uikit/ui_nib.rs
```diff
@@ -170,8 +170,15 @@ pub const CLASSES: ClassExports = objc_classes! {
...
 ///                     optionsOrNil:nil];
 /// ```
-pub fn load_main_nib_file(env: &mut Environment, _ui_application: id) {
-    let path = env.bundle.main_nib_file_path();
+/// Returns true if the main nib file was loaded, false if not.
+pub fn load_main_nib_file(env: &mut Environment, _ui_application: id) -> bool {
+    let path = match env.bundle.main_nib_file_path() {
+        Some(path) => path,
+        None => {
+            // The application doesn't have a main nib file.
```
hikari-no-yume, line 178: Strangely enough, this isn't sufficient. The newest version of Crash Bandicoot Nitro Kart 3D has an NSMainNibFile key but no actual file by that name. I have no idea why this works.

I already wrote code to handle that case, so maybe I should just clean up and push what I have? Sorry about the duplicated effort.

hikari-no-yume: (Perhaps the main nib file is only loaded if a delegate class name isn't specified…? The documentation isn't clear.)

src/frameworks/uikit/ui_view.rs
```diff
@@ -67,7 +67,26 @@ pub const CLASSES: ClassExports = objc_classes! {
     env.objc.get_known_class("CALayer", &mut env.mem)
 }
 
-// TODO: initWithFrame:, accessors, etc
+- (id)initWithFrame:(CGRect)frame { // CGRect
+    let host_object: &mut UIViewHostObject = env.objc.borrow_mut(this);
+    host_object.bounds = frame;
+    host_object.center = CGPoint { x: frame.size.width / 2.0, y: frame.size.height / 2.0 };
```
hikari-no-yume, line 73: FYI, from what I remember of the docs for `bounds`, `center` and `frame`, I think this is wrong:

* The center will have to be offset by the origin in the frame also
* The origin in the bounds is probably 0, at the very least it's not the same as the origin of the frame

I also already have some code to handle this, so I can clean up and push that instead if you want.

Though on that note, I'm not perfectly sure my own understanding is correct.

src/frameworks/uikit/ui_view.rs
```diff
@@ -109,6 +128,12 @@ pub const CLASSES: ClassExports = objc_classes! {
     // TODO: enable multitouch
 }
 
+- (())addSubview:(id)view { // UIView *
+    () = msg![env; view layoutSubviews];
+    // TODO: keep a list of subviews somewhere
+    () = msg![env; this layoutSubviews];
```
hikari-no-yume, line 134: I'm guessing you forgot to delete one of these?

hikari-no-yume: Oh wait sorry I didn't notice the two different receivers.

hikari-no-yume:
I cherry-picked two of your commits (the icon file and `makeKeyAndVisible` ones) and for the other commits, I cleaned up and pushed my own code that does similar things instead (eaa5fab36fd6cc8e9785fa2e0c6efcae532523b2, e15f679244d5ca16baa1089a1e534a6960edbd0f, d208e2f43bea53aca18c61da969a2e01c0b0a06d, d2239e94a28c92c5ff0bf1a48cb12a4d32338acd). The last of those takes some inspiration from one of your commits. Sorry about the duplicate effort…

By the way, I fixed the formatting of your message for the icon file commit.

nitinseshadri:
Thanks!

(Was busy over the past few days and just saw this, would have replied to your code review comments otherwise)

Issue #10, opened 2023-02-04
Not all Info.plist files have a CFBundleName (thread 'main' panicked at 'IndexMap: key not found', src\bundle.rs:50:27)


ghost:
```
PS D:\다운로드\touchHLE_v0.1.0_Windows_x86_64> .\touchHLE.exe "OvenBreak.app"
touchHLE v0.1.0 — https://touchhle.org/

thread 'main' panicked at 'IndexMap: key not found', src\bundle.rs:50:27
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
This happen when I try to run Ovenbreak (paid version) 1.09 ~ 1.12 ipa.

```
PS D:\다운로드\touchHLE_v0.1.0_Windows_x86_64> .\touchHLE.exe "OvenBreak.app"
touchHLE v0.1.0 — https://touchhle.org/

Error: "Could not parse icon image"
```
As an aside, this error happen when I try to run Ovenbreak (paid version) 1.0 ~ 1.08 ipa, but seems like it already commented on other thread.

hikari-no-yume:
Oh right, some app bundles don't have the CFBundleName key in their Info.plist file, I should fix that eventually.

To be clear, exactly one app (Super Monkey Ball) is expected to work right now, but I'll keep this open as a reminder of this particular issue. It's extremely likely that there are dozens or hundreds of other issues that would prevent OvenBreak from working.

hikari-no-yume:
I fixed this recently in c1c746202b499f1b1c9c3b2acd04bf7de7968637 but forgot about this issue.

PR #65, opened 2023-02-22
fix for bundles that don't use CFBundleName (issue #10)

dansl:
fix for bundles that don't use CFBundleName (issue #10)

### DIFF
--- a/src/bundle.rs
+++ b/src/bundle.rs
@@ -35,7 +35,12 @@ impl Bundle {
             .into_dictionary()
             .ok_or_else(|| "plist root value is not a dictionary".to_string())?;
 
-        let bundle_name = plist["CFBundleName"].as_string().unwrap();
+        let mut bundle_name = "";
+        if plist.contains_key("CFBundleName"){
+            bundle_name = plist["CFBundleName"].as_string().unwrap();
+        }else if plist.contains_key("CFBundleDisplayName"){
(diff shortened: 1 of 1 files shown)


src/bundle.rs
```diff
@@ -35,7 +35,10 @@ impl Bundle {
             .into_dictionary()
             .ok_or_else(|| "plist root value is not a dictionary".to_string())?;
 
-        let bundle_name = plist["CFBundleName"].as_string().unwrap();
+        let mut bundle_name = "";
+        if plist.contains_key("CFBundleName") {
+            bundle_name = plist["CFBundleName"].as_string().unwrap();
+        }
```
hikari-no-yume, line 41: This can be condensed to:

```rust
let bundle_name = plist.get("CFBundleName").map_or("", |s| s.as_string().unwrap());
```

But I don't think using an empty string is a good solution, we should use the filename the bundle already has in that case.

dansl: 👍 Good to know. This is my first time using Rust, so still learning a lot about the language. Wasn't even sure if I could do something shorthand like `let bundle_name = plist.contains_key("CFBundleName") ? plist["CFBundleName"].as_string().unwrap() : "";`

hikari-no-yume:
FYI I've finally fixed this: c1c746202b499f1b1c9c3b2acd04bf7de7968637.

PR #66, opened 2023-02-22
Fix reminent when app support was a readme section

upintheairsheep:


hikari-no-yume:
Thanks for catching this! :)

PR #69, opened 2023-02-25
Added basic initial wchar function wctob.

TylerJaacks:
So I added a very basic implementation of the wctob function from wchar.h. This function does not do the conversion, but should get past the error. I don't no what binary had this function so I can't test it. Also this is like my first Rust project so please be nice :).

### DIFF
--- a/src/libc.rs
+++ b/src/libc.rs
@@ -23,6 +23,7 @@ pub mod stdio;
 pub mod stdlib;
 pub mod string;
 pub mod time;
+pub mod wchar;
 
 /// Container for state of various child modules
 #[derive(Default)]
@@ -34,4 +35,5 @@ pub struct State {
     stdlib: stdlib::State,
     string: string::State,
     time: time::State,
+    wchar: wchar::State,
 }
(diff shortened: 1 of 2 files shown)


hikari-no-yume:
Hi, I appreciate the effort but I'm going to have to reject this for a number of reasons.

By your own words, you haven't tested it, and the implementation “does not do the conversion”. I also suspect you haven't even compiled it, because there are problems with it that the Rust compiler would have warned you about. I also notice that you @'d someone in an existing issue and asked them to test it. I think that's appropriate in cases where what someone is doing is difficult to reproduce, but with something simple like this, I don't think the burden should be offloaded to the issue reporter.

I also don't like that this copies from glibc. It is at least attributed, and glibc is under a license compatible with touchHLE's release builds, so there isn't a copyright infringement. But this would be the first non-MPL-licensed code in the project outside a dependency, and I think that demands a justification, and that's actually my main problem with this: there's no obvious reason glibc _should_ be used as a reference. iPhone OS doesn't use glibc, so this doesn't bring us closer to the real implementation's behaviour, and this function seems to be simple enough that the documentation should be enough.

hikari-no-yume:
Also, please read [the contributing guide](https://github.com/hikari-no-yume/touchHLE/blob/trunk/CONTRIBUTING.md).

PR #78, opened 2023-03-08
Implement more UIDevice calls

upintheairsheep:
Suggestions/TODO: 
1. Make most or every field customizable via flags (like the version)
2. Implement all Battery related functions, redirecting them from the host machine. 
3. Rename the device name to the host machine’s user set name or computer model name?

### DIFF
--- a/src/frameworks/uikit/ui_device.rs
+++ b/src/frameworks/uikit/ui_device.rs
@@ -60,8 +60,8 @@ pub const CLASSES: ClassExports = objc_classes! {
     ns_string::get_static_str(env, "UIUserInterfaceIdiom.phone")
 }
 
-- (id)systemName {
-    ns_string::get_static_str(env, "iPhoneOS")
+- (id)name {
+    ns_string::get_static_str(env, "touchHLE")
 }
 
 
--- a/src/frameworks/uikit/ui_device.rs
+++ b/src/frameworks/uikit/ui_device.rs
@@ -47,17 +47,24 @@ pub const CLASSES: ClassExports = objc_classes! {
     }
 }
 
-- (())beginGeneratingDeviceOrientationNotifications {
-    log!("TODO: beginGeneratingDeviceOrientationNotifications");
-}
-- (())endGeneratingDeviceOrientationNotifications {
-    log!("TODO: endGeneratingDeviceOrientationNotifications");
-}
-
 // NSString
 - (id)systemVersion {
     ns_string::get_static_str(env, "2.0")
 }
+
+- (id)systemName {
+    ns_string::get_static_str(env, "iPhoneOS")
+}
+
+- (id)UI_USER_INTERFACE_IDIOM {
+    ns_string::get_static_str(env, "UIUserInterfaceIdiom.phone")
+}
+
+- (id)systemName {
+    ns_string::get_static_str(env, "iPhoneOS")
+}
+
+
 
 - (bool)isMultitaskingSupported {
     false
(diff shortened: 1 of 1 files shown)


src/frameworks/uikit/ui_device.rs
```diff
@@ -47,18 +47,25 @@ pub const CLASSES: ClassExports = objc_classes! {
...
 
+- (id)systemName {
+    ns_string::get_static_str(env, "iPhoneOS")
+}
+
+- (id)UI_USER_INTERFACE_IDIOM {
+    ns_string::get_static_str(env, "UIUserInterfaceIdiom.phone")
+}
+
+- (id)systemName {
```
hikari-no-yume: You didn't mean to define this twice, right?

src/frameworks/uikit/ui_device.rs
```diff
@@ -47,18 +47,25 @@ pub const CLASSES: ClassExports = objc_classes! {
...
 // NSString
 - (id)systemVersion {
     ns_string::get_static_str(env, "2.0")
 }
 
+- (id)systemName {
+    ns_string::get_static_str(env, "iPhoneOS")
+}
+
+- (id)UI_USER_INTERFACE_IDIOM {
```
hikari-no-yume: I don't think this is meant to be an instance method. According to the docs it's a function. Did you miss that or am I missing something?

src/frameworks/uikit/ui_device.rs
```diff
@@ -47,18 +47,25 @@ pub const CLASSES: ClassExports = objc_classes! {
     }
 }
 
-- (())beginGeneratingDeviceOrientationNotifications {
```
hikari-no-yume, line 50: Why were these removed?

hikari-no-yume:
> 1\. Make most or every field customizable via flags (like the version)

Hmm, I don't think there's going to be much practical use in being able to customize most of these? I'll definitely need to add a flag eventually that changes whether touchHLE pretends to be an iPhone/iPod touch or an iPad, though, once we have iPad apps working.

> 2\. Implement all Battery related functions, redirecting them from the host machine.

Oh yes, this would be cute. Relatedly, something I'm going to do eventually is implement an iOS-style status bar at the top of the screen (for apps which don't hide it).

> 3\. Rename the device name to the host machine’s user set name or computer model name?

I guess I could do this but it doesn't sound useful exactly.

hikari-no-yume (changes requested):
See comments.

src/frameworks/uikit/ui_device.rs
upintheairsheep: Sorry

hikari-no-yume:
Have you tested this, and in what ways?

upintheairsheep:
> Have you tested this, and in what ways?

No I am physically not capable of testing this. All thanks to funkin' chromeOS. I have a Windows device, but I made a grave mistake while buying one and had to get a 5400 RPM hard drive, which is basically unusable. I am feeling regret with myself while typing.

upintheairsheep:
By the way, which version should be that last version I implement calls/functions from? iOS 4? iOS 5? iOS 6? iOS 7?

hikari-no-yume:
Hmm, I know you mean well but I really don't want to accept untested code on principle.

If you're on Chrome OS, I'm pretty sure it's possible to install Linux somehow, which would let you get a reasonable environment for development.

> By the way, which version should be that last version I implement calls/functions from? iOS 4? iOS 5? iOS 6? iOS 7?

There's not really much point in implementing stuff for versions we don't support yet. I haven't implemented anything that's not in iPhone OS 2 because I can't actually test it.

In general, also, I don't want to implement things that can't be tested yet unless they're extremely simple and obviously correct. There's a risk that untested code turns out to be useless or, worse, breaks something in future. Almost all the existing code in touchHLE is used by one of the supported apps, and the rest is mostly either extremely simple and closely related to code that _is_ tested, or is used by an unsupported app and has been tested against that.

PR #88, opened 2023-03-21
[WIP] Android build

ciciplusplus:
# What is this all about?

This is a **WIP** Android build, supporting aarch64 devices

# What was tested?

- ~~Super Monkey Ball 1.02 on Google Pixel 3a (playable, but tilt controls not working, so not _that_ playable)~~ All supported apps (fully playable!)
- touchHLE build for macOS (builds, runs)

I've not tested Windows version after the changes

# How to use

Refer to BUILDING.md in android folder for building instruction.

**IMPORTANT**: in the current state Android app expects a particular setup for file folder to work.
In particular, `/data/data/org.touchhle.android/files` folder on your device should contain:
- ~~`Super Monkey Ball  v1.02 .ipa` (legally obtained and owned; _note 2 spaces between Ball and v1.02_)~~ a `touchHLE_apps` folder with your apps
- `touchHLE_fonts` and `touchHLE_dylibs` folders from touchHLE source or binary distribution

# Important changes

- Introduction of android project, which is a copy of SDL2 2.26.4 android template project with small modifications (zlib LICENSE.txt file, cargo-ndk plugin and some cosmetic changes)
- ~~main.rs is split now to lib.rs (android build) and main.rs (desktop build)~~
- passing from [bundled, static-link] mode to [bundled] for sdl2 dependency
- ~introducing OpenGL ES 1.1 native implementation~
- Android app has `org.touchhle.android` bundle
- SDL source is added to vendor/ folder, used only for Android build

# Known issues

- ~~No mechanism for app/ipa selection/loading~~
- ~Tilt controls are not implemented~
- Splash screen is disabled
- ~App scale/orientation is hardcoded~
- Only arm64-v8a arch is supported
- ~~SDL is built twice: once with rust-sdl2 and once specifically for Android~~
- SDL 2.26.6 version was patched [here](https://github.com/ciciplusplus/SDL/commit/6c57af8b3e2f571e1d478086dd77f7ca9b7281cd) (and as consequence sdl2-rust as well) to workaround SDL build issue with ANDROID_NDK not set... Something is wrong here on one of the crate among cmake, cargo-ndk, sdl2-rust or cargo itself, as locally I can do build SDL from sources (and it's also done during gradle build). Maybe it's related to the changes of Android tooling itself. Or maybe it's cmake config of SDL2 which is wrong. Or...
- ~~Some GL functions in update_renderbuffers do not have equivalents in Gles 1.1, so i just commented them out. While I've not observed any visible glitches, the cleaning/restoring of gl context may not work perfectly here.~~
- ~~Logging with println!() is not working. Normally, with android_log it should be redirected to logcat, but, firstly, `log` dependency of it conflicts with internal log module and, secondly, even if fixed, it's still not working. At the brighter side, sdl2::log works, so all log!() calls using it instead~~

# References

1. https://julhe.github.io/posts/building-an-android-app-with-rust-and-sdl2/
2. https://github.com/libsdl-org/SDL/blob/release-2.26.4/docs/README-android.md
3. https://github.com/rust-mobile/rust-android-examples/tree/main/na-mainloop

### P.S.
Hope that my monkey patching (pun intended) it's not too big to review 😸


hikari-no-yume:
Congratulations on getting this in a good enough state to open a pull request? And thanks for your efforts on this, I know it's been a lot of work!

To others: please do not comment on this PR asking when this will be merged, thanks in advance.

ciciplusplus:
> Congratulations on getting this in a good enough state to open a pull request? And thanks for your efforts on this, I know it's been a lot of work!

Thanks and sorry, in case it wasn't clear, I think it's ready to be reviewed.

It would be great to receive at least high level feedback, like which parts are absolutely necessary to be reworked and which are ok to merge in the current state.

src/cpu/dynarmic_wrapper/build.rs
```diff
@@ -57,6 +58,14 @@ fn main() {
         }
         build.define("Boost_INCLUDE_DIR", boost_path);
     }
+    let os = env::var("CARGO_CFG_TARGET_OS").expect("CARGO_CFG_TARGET_OS was not set");
+    if cfg!(target_os = "macos") && os.eq_ignore_ascii_case("android") {
+        // TODO: how to set version independent path?
```
hikari-no-yume: There's already a special solution for this on Windows, maybe you can just re-use this for Android and have the user set up a symlink? It's ugly of course, but it avoids hardcoding the path.

src/frameworks/opengles.rs
```diff
@@ -28,11 +28,16 @@
 //!   - [EXT_framebuffer_object](https://registry.khronos.org/OpenGL/extensions/EXT/EXT_framebuffer_object.txt)
 
 pub mod eagl;
+mod gles1_native;
 mod gles1_on_gl2;
 mod gles_generic;
 mod gles_guest;
 
+#[cfg(target_os = "android")]
```
hikari-no-yume: Hmm, I'd like it if native OpenGL ES is available as an option on desktop OSes that support it, so can we remove the conditional compilation here? Linux with Mesa drivers supports it so it should just work on that platform. Apparently, some vendors' drivers on Windows might have weird extensions that also provide this…

src/frameworks/opengles/eagl.rs
```diff
@@ -353,3 +369,168 @@ unsafe fn present_renderbuffer(env: &mut Environment) {
 
     //{ let err = gl::GetError(); if err != 0 { panic!("{:#x}", err); } }
 }
+
+/// Variant of present_renderbuffer but for Android
+/// Instead of blitting between framebuffers, we use CopyTexImage2D to copy renderbuffers to a texture
+/// TODO: refactor common parts with present_renderbuffer
+#[cfg(target_os = "android")]
```
hikari-no-yume: I'm not sure how realistic this is, but can we share most of the code rather than duplicating the whole function?

src/fs.rs
```diff
@@ -406,7 +407,16 @@ impl Fs {
 
         let bundle_guest_path = home_directory.join(&bundle_dir_name);
 
-        let documents_host_path = Path::new("touchHLE_sandbox")
+        // TODO: instead of path hardcoding, we should use https://wiki.libsdl.org/SDL2/SDL_AndroidGetInternalStoragePath
+        // but... it's not exposed via rust-sdl2 :(
```
hikari-no-yume: Can we have a single function somewhere for getting the path? It can be shared between this file and the file that loads fonts. I guess it might also be used for loading the default settings at some point.

If we want to use `SDL_AndroidGetInternalStoragePath` in that function, we can probably access it via the rust sdl2-sys crate.

android/app/src/main/java/org/libsdl/app/HIDDeviceBLESteamController.java
hikari-no-yume, line 1: Seems like a lot of code that is just coming from SDL and we wouldn't want to modify. Can we automatically fetch it from the SDL repo or something like that?

.gitignore
```diff
@@ -1,3 +1,19 @@
...
 /vendor/boost
 /touchHLE_sandbox
+
+# Android
+
+*.iml
+.gradle
+/local.properties
+.idea/
+.DS_Store
```
hikari-no-yume: `.DS_Store` isn't Android-specific. I recommend using a global git “excludes file” so you don't have to put this in a `.gitignore` file in every repository.

.gitignore
```diff
@@ -1,3 +1,19 @@
...
+/build
+/captures
+.externalNativeBuild
+.cxx
+local.properties
+*.so
```
hikari-no-yume: Huh, why is there a wildcard ignore for `.so` files, surely they won't be dumped at the base of the repository?

Cargo.toml
```diff
@@ -36,7 +40,7 @@ rusttype = "0.9.3"
 # which is old enough that its bundled SDL2 version is missing various nice
 # things, and doesn't build on macOS.
 # This is a temporary fork to get a newer SDL2 version.
-sdl2 = { git = "https://github.com/hikari-no-yume/rust-sdl2.git", branch = "update-bundled-SDL2", features = ["bundled", "static-link"] }
+sdl2 = { git = "https://github.com/ciciplusplus/rust-sdl2.git", rev = "a5900b4acd55facebb75cfcd6ca5691543111487", features = ["bundled"] }
```
hikari-no-yume: This is a different revision to the submodule? Why?

android/app/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,100 @@
...
+     com.gamemaker.game
+-->
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="org.touch.hle"
+    android:versionCode="1"
+    android:versionName="1.0"
+    android:installLocation="auto">
+
+    <!-- OpenGL ES 2.0 -->
+    <uses-feature android:glEsVersion="0x00020000" />
```
hikari-no-yume: I guess it doesn't matter, but this is wrong, right? :)

android/gradlew
hikari-no-yume, line 1: Does Android not provide some global script so this doesn't have to be in the repo…?

hikari-no-yume:
I had a first look at the code, finally. It looks like the changes to the main touchHLE code were pretty minor, which is encouraging. I don't like how many new files are needed just to deal with Android build-system stuff though. I'll try to get an Android development environment set up this week. I'm hoping I might be able to find a way to simplify it.

hikari-no-yume:
Nitpick: the website is `touchhle.org` so the package name should be `org.touchhle`. (Edit: maybe `org.touchhle.android`? idk)

Cargo.toml
ciciplusplus: SDL 2.26.6 version was patched [here](https://github.com/ciciplusplus/SDL/commit/6c57af8b3e2f571e1d478086dd77f7ca9b7281cd) (and as consequence sdl2-rust as well) to workaround SDL build issue with ANDROID_NDK not set... Something is wrong here on one of the crate among cmake, cargo-ndk, sdl2-rust or cargo itself, as locally I can do build SDL from sources (and it's also done during gradle build). Maybe it's related to the changes of Android tooling itself. Or maybe it's cmake config of SDL2 which is wrong. Or...

src/frameworks/opengles/eagl.rs
ciciplusplus: Well, I see 2 possibilities:

- use present_renderbuffer's platform specific versions
- have single function filled with platform specific branching inside function body

I opted for first one for the sake of readability, but can try to implement second option if you think it's preferable

android/app/src/main/java/org/libsdl/app/HIDDeviceBLESteamController.java
ciciplusplus: I can try to symlink them to the project, but this comes with down sides:

1) If SDL project changes layout, those symlinks need to be updated
2) Eventually, I think we should get rid of SDL repo as dependency, so creating more reasons to keep it around is not that great.

IMHO this will make Andoroid project build more fragile, but please let me know if it's a preferable way for you

.gitignore
ciciplusplus: i followed SDL .gitgignore https://github.com/libsdl-org/SDL/blob/release-2.26.4/.gitignore#L20

Otherwise generated .so are in `android/app/src/main/jniLibs/arm64-v8a/libtouchHLE.so`, so i can git-ignore `android/app/src/main/jniLibs` instead

android/gradlew
ciciplusplus: Gradle build system (used by Android) recommends to use Gradle wrapper.
From their docs https://docs.gradle.org/current/userguide/gradle_wrapper.html

> The recommended way to execute any Gradle build is with the help of the Gradle Wrapper (in short just “Wrapper”). The Wrapper is a script that invokes a declared version of Gradle, downloading it beforehand if necessary. As a result, developers can get up and running with a Gradle project quickly without having to follow manual installation processes saving your company time and money.

i would advise to keep it ;)

src/frameworks/opengles.rs
ciciplusplus: Done
(but I need to add unused_imports to silence a warning)

android/app/src/main/AndroidManifest.xml
ciciplusplus: yeap, the whole SDL project was copy-pasted (as advised for "complex" projects), i didn't saw that one

Removed

.gitignore
ciciplusplus: Done

.gitignore
ciciplusplus: Done

src/cpu/dynarmic_wrapper/build.rs
ciciplusplus: Done

src/fs.rs
ciciplusplus: Done for a common function, i placed it as pub in the fs module

android/app/src/main/java/org/libsdl/app/HIDDeviceBLESteamController.java
ciciplusplus: I changed to symlinking them

ciciplusplus:
> Nitpick: the website is `touchhle.org` so the package name should be `org.touchhle`. (Edit: maybe `org.touchhle.android`? idk)

Done, changed to `org.touchhle.android`

upintheairsheep:
Suggestions: /data/data/org.touch.hle/files should be /sdcard/Android/data/org.touchhle.android/files otherwise root or android backup hacking would be required to use touchHLE on Android, which as rooting gets less and less popular, a large inconvenience. We also should add a basic GUI like J2ME Loader and Dolphin does, so you have way more freedom and accessibility.

src/frameworks/opengles.rs
hikari-no-yume: Oh, because the cfg check still exists in other places? I was imagining you could remove all the `cfg(target_os = "android")` checks related to GLES1Native. Instead of always using GLES1Native on Android and GLES1OnGL2 on desktop, we can select dynamically. But given that GLES1OnGL2 is more mature, we should make it the default for now. (Ask me on Discord if you want me to elaborate. Also, if it's more annoying than expected, maybe we can skip it for now.)

src/frameworks/opengles/eagl.rs
hikari-no-yume: Oh wait I forgot that it bypasses the API abstraction, so it is actually API-specific right now :(

I guess this conflicts with what I said above about wanting GLES1Native to be available on desktop. Okay, I guess we'll leave it as-is for now and we can clean it up later.

src/frameworks/opengles.rs
hikari-no-yume: Actually never mind, considering the presentRenderbuffer thing (below), this isn't going to be practical for now. You can put back the cfg check.

android/app/src/main/java/org/libsdl/app/HIDDeviceBLESteamController.java
hikari-no-yume: Thanks!

> If SDL project changes layout, those symlinks need to be updated

This is true, but wouldn't there be the same risk that the Java files have to be updated when SDL is updated?

> Eventually, I think we should get rid of SDL repo as dependency, so creating more reasons to keep it around is not that great.

Yeah it's not ideal to have a new submodule. The one alternative I can see is having some script that will download the SDL release and copy the relevant files. That would avoid symlink problems on Windows, but we'd have to support all OSes for the script, so… I think symlinks are fine for now.

src/cpu/dynarmic_wrapper/build.rs
hikari-no-yume: Hmm this condition is a bit weird, shouldn't it be `os.eq_ignore_ascii_case("windows") || os.eq_ignore_ascii_case("android")`?

src/fs.rs
hikari-no-yume: Thanks, makes sense to me :)

src/cpu/dynarmic_wrapper/build.rs
ciciplusplus: 😅 you're right, done

src/frameworks/opengles.rs
ciciplusplus: Done 😃

Cargo.toml
ciciplusplus: Regarding this, i think if you could confirm that it works/builds on your side without changing revision, i'll remove this part from PR and only keep the removal of "static-link"

(i would still need to keep revision change just for me, but this is not a big deal)

Boom20230908:
I would like to hire you to work on a challenging project with a minimum budget of $10W

Please contact me

https://t.me/BoomCashs

CPAdfvhh (approved):

Issue #86, opened 2023-03-16
Move to dynamic linking of SDL2


hikari-no-yume:
touchHLE currently uses static linking for SDL2.

The SDL project makes a convincing moral argument that SDL2 shouldn't be statically linked, because that prevents users from running old software with an updated library to improve compatibility. As a game preservation enthusiast, I can't argue against that! On the other hand, SDL 2 actually can be [dynamically replaced _even when statically linked_](https://github.com/libsdl-org/SDL/blob/SDL2/docs/README-dynapi.md), so maybe it doesn't matter so much.

The reason I use static linking in touchHLE is that it makes it easier to redistribute the result, in theory: it's a self-contained EXE file! Except not really, because you need the `touchHLE_dylibs` and `touchHLE_fonts` directories. So an extra file (the dynamic library for SDL2) wouldn't really make things worse.

It would also solve some of our current problems, because statically linking to SDL2 relies on using the `bundled` feature of the `sdl2` crate. This has left us [stuck with old SDL2 versions](https://github.com/hikari-no-yume/touchHLE/issues/46) until the Rust crate makes a new release or we [make a custom fork](https://github.com/hikari-no-yume/rust-sdl2/tree/update-bundled-SDL2). Ironically, it also causes build issues: while it makes it easier to build on typical Windows and macOS setups, I gather it'd be easier to build for macOS (when cross-compiling), Linux and Android if you could just dynamically link a pre-compiled SDL2.

So: at some point, we should stop using the `static` and `bundled` features for the `sdl2` crate. That's trivial to do. The only slightly tricky thing is release packaging. Will we have to start using an app bundle on macOS, for example?

PR #90, opened 2023-03-23
Implement dynamic linking & Fix Linux building

GeffDev:
This PR implements a dynamic linking compile time feature, which can be enabled with `--features dynamic`, or in the `Cargo.toml`. This PR also fixes building on Linux. Tested on Linux, Windows and MacOS (Big Sur).

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -10,7 +10,7 @@ There is one special external dependency, Boost:
 * On other OSes, install Boost from your package manager. If you are on macOS and using [Homebrew](https://brew.sh/): `brew install boost`.
 
 Then you just need to run `cargo run --release` (for a release build) or `cargo run` (for a debug build) to build and run touchHLE. On an underpowered, passively-cooled, 2-core laptop (2017 Retina MacBook), a clean release build takes a bit less than 9 minutes.
-* To compile with dynamically linked libraries, append `--features dynamic` to the end of the cargo build command.
+* To compile with dynamically linked libraries, append `--no-default-features --features dynamic` to the end of the cargo build command.
 
 The `touchHLE_dylibs` and `touchHLE_fonts` directories contain files that the resulting binary will need at runtime, so you'll need to copy them if you want to distribute the result. You also should include the license files.
 
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -10,7 +10,8 @@ There is one special external dependency, Boost:
 * On other OSes, install Boost from your package manager. If you are on macOS and using [Homebrew](https://brew.sh/): `brew install boost`.
 
 Then you just need to run `cargo run --release` (for a release build) or `cargo run` (for a debug build) to build and run touchHLE. On an underpowered, passively-cooled, 2-core laptop (2017 Retina MacBook), a clean release build takes a bit less than 9 minutes.
-* To compile with dynamically linked libraries, append `--no-default-features` to the end of the cargo build command.
+
+touchHLE can also be dynamically linked (which means instead of using the bundled dependencies, it will use the dependencies provided by your system). To build a dynamically linked version of touchHLE, you will need to have the SDL2 and OpenAL shared libraries installed, and then you can append `--no-default-features` (this flag is passed in to disable static linking, which is the default) to the end of the cargo build command.
 
 The `touchHLE_dylibs` and `touchHLE_fonts` directories contain files that the resulting binary will need at runtime, so you'll need to copy them if you want to distribute the result. You also should include the license files.
 
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -10,7 +10,7 @@ There is one special external dependency, Boost:
 * On other OSes, install Boost from your package manager. If you are on macOS and using [Homebrew](https://brew.sh/): `brew install boost`.
 
 Then you just need to run `cargo run --release` (for a release build) or `cargo run` (for a debug build) to build and run touchHLE. On an underpowered, passively-cooled, 2-core laptop (2017 Retina MacBook), a clean release build takes a bit less than 9 minutes.
-* To compile with dynamically linked libraries, append `--no-default-features --features dynamic` to the end of the cargo build command.
+* To compile with dynamically linked libraries, append `--no-default-features` to the end of the cargo build command.
 
 The `touchHLE_dylibs` and `touchHLE_fonts` directories contain files that the resulting binary will need at runtime, so you'll need to copy them if you want to distribute the result. You also should include the license files.
 
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -10,6 +10,7 @@ There is one special external dependency, Boost:
 * On other OSes, install Boost from your package manager. If you are on macOS and using [Homebrew](https://brew.sh/): `brew install boost`.
 
 Then you just need to run `cargo run --release` (for a release build) or `cargo run` (for a debug build) to build and run touchHLE. On an underpowered, passively-cooled, 2-core laptop (2017 Retina MacBook), a clean release build takes a bit less than 9 minutes.
+* To compile with dynamically linked libraries, append `--features dynamic` to the end of the cargo build command.
 
 The `touchHLE_dylibs` and `touchHLE_fonts` directories contain files that the resulting binary will need at runtime, so you'll need to copy them if you want to distribute the result. You also should include the license files.
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,10 +19,8 @@ authors = { workspace = true }
 homepage = { workspace = true }
 
 [features]
-# Static linking is default.
 default = ["static"]
-dynamic = ["sdl2/use-pkgconfig"]
-static = ["sdl2/bundled", "sdl2/static-link"]
+static = ["sdl2/bundled", "sdl2/static-link", "touchHLE_openal_soft_wrapper/static", "touchHLE_dynarmic_wrapper/static"]
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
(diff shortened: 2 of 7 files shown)


hikari-no-yume:
Ah, thanks, this has been on my TODO list since even before release 0.1.0 (since another Linux user wanted it), but I never got round to it. This looks good at first glance. I'll look closer tomorrow but I'd expect this can be merged quickly.

hikari-no-yume:
Hmm… does it actually make sense to just switch to dynamic linking like this? As I perceive it, the benefit of using dynamic linking is you can use your distribution's versions of certain libraries (OpenAL Soft, SDL2, maybe some of dynarmic's dependencies but that feels like a stretch). But if that's the goal, why also build our own versions of the dependencies?

GeffDev:
Yeah, using the distribution's libraries would work. I see Rust-SDL2 has a `use-pkgconfig` feature, so we could use that to dynamically link SDL2. At the moment, the only libraries I can think of that we could use from the distribution would be OpenAL Soft and SDL2. I'll check this out tonight.

Another thing about the SDL2 crate would be that to dynamically link it, the user would manually have to edit the `Cargo.toml` to change the default `static-link` and `bundled` feature in the SDL2 crate to `use-pkgconfig`, because afaik you can't pass features into sub-crates? Anyway, I'll have a look.

hikari-no-yume:
FWIW I am considering moving to using dynamic linking only for SDL2: https://github.com/hikari-no-yume/touchHLE/issues/86.

For Windows and macOS I'd bundle one of the official release binaries, and then for Linux users it'd probably “just work” if you have SDL2 installed from your distro's package manager.

hikari-no-yume:
Hmm, looks like we can [enable features based on features](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features).

(Edit: sorry, I initially quoted the wrong bit there…)

GeffDev:
Huh, nice! I used that for the static and dynamic features. From what I can see, this should now use the system libraries for SDL2 and OpenAL if the `dynamic` feature is defined.

I see you mentioned #86 earlier, when you plan on switching to that, you can change the `default` feature from `static` to `dynamic`, not sure if you plan on migrating over to dynamic linking for OpenAL too though. 

I dunno much about how shared libraries work on Windows and MacOS, but I'm presuming it's pretty much the same across all platforms (add to PATH). This could be a bit more documented in `BUILDING.md` with what libraries have to be installed and how to install them, but :shrug:.

src/audio/openal_soft_wrapper/build.rs
```diff
@@ -12,36 +12,48 @@ fn link_search(path: &Path) {
...
-    let openal_soft_out = build.build();
+        let mut build = cmake::Config::new(workspace_root.join("vendor/openal-soft"));
+        build.define("LIBTYPE", "STATIC");
+        let openal_soft_out = build.build();
 
+        link_search(&openal_soft_out.join("lib"));
+        // some Linux systems
+        link_search(&openal_soft_out.join("lib64"));
+    }
     // Some dependencies of OpenAL Soft.
```
hikari-no-yume: This should also be skipped I think, it's probably specific to static linking (I assume the dynamic library links against everything it needs to already)

hikari-no-yume: Or wait, am I wrong? I see you link sndio on Linux below…

src/cpu/dynarmic_wrapper/build.rs
```diff
@@ -67,11 +73,16 @@ fn main() {
...
+    if !cfg!(target_os = "linux") {
+        link_lib(if cfg!(debug_assertions) {
+            "fmtd"
+        } else {
+            "fmt"
+        });
     } else {
-        "fmt"
-    });
+        println!("cargo:rustc-link-lib=dylib=fmt");
```
hikari-no-yume: Does fmt have to be dynamic? Does that actually work?

Cargo.toml
```diff
@@ -18,6 +18,12 @@ license = { workspace = true }
 authors = { workspace = true }
 homepage = { workspace = true }
 
+[features]
+# Static linking is default.
+default = ["static"]
+dynamic = ["sdl2/use-pkgconfig"]
+static = ["sdl2/bundled", "sdl2/static-link"]
```
hikari-no-yume: Hmm, here you have distinct “dynamic” and “static” features, but in one of the build.rs you're doing an if-not-dynamic check as if that implies static, but that's not guaranteed right? I think it'd be cleaner to just have a “static” feature, and if it's not present, then assume dynamic.

Also, maybe there could be separate “SDL2 static” and “OpenAL static” features? (By the way, wouldn't you have to update the Cargo.toml files for the inner crates for the static linking feature to pass through? Did you test the current setup actually works?)

src/audio/openal_soft_wrapper/build.rs
GeffDev: Yeah, you're right, `sndio` only needs to be linked when building statically. Not sure if that is the case with frameworks (dunno much about Mac), but I'll move it in the static block anyway.

Cargo.toml
GeffDev: Yeah, that's a good idea. I removed the dynamic feature in my commit I'm about to push. I made a static feature in the `openal_soft` crate (and the `dynarmic_wrapper` crate, not sure if it's used there), and the static feature enables them by default. 

If you only need to dynamically link SDL2 and statically link OpenAL, you can just do something like `cargo build --no-default-features --features touchHLE_openal_soft_wrapper/static`.

src/cpu/dynarmic_wrapper/build.rs
GeffDev: Yup, there is no static library for fmt, only a shared library. Same with sndio.

hikari-no-yume: Oh I have a guess as to what's going on: maybe we need to explicitly set the `DYNARMIC_USE_BUNDLED_EXTERNALS` cmake flag. On my macOS and Windows machines, I guess I just don't have any of those external libraries installed system-wide, but you have some of them, so that this works at all is an accident of our build environments.

hikari-no-yume: Specifically, what I think is going on is:

* On my machines, I have none of those libraries installed system-wide, so dynarmic falls back to using its bundled versions and the static libraries need to be linked.
* On your machine, you have fmt installed system-wide, so dynarmic doesn't build its bundled static version of _that_ library, but does build them for the rest.

I've been assuming this bundled approach the whole time, but I didn't realise it was environment-dependent.

BUILDING.md
```diff
@@ -10,6 +10,7 @@ There is one special external dependency, Boost:
 * On other OSes, install Boost from your package manager. If you are on macOS and using [Homebrew](https://brew.sh/): `brew install boost`.
 
 Then you just need to run `cargo run --release` (for a release build) or `cargo run` (for a debug build) to build and run touchHLE. On an underpowered, passively-cooled, 2-core laptop (2017 Retina MacBook), a clean release build takes a bit less than 9 minutes.
+* To compile with dynamically linked libraries, append `--no-default-features` to the end of the cargo build command.
```
hikari-no-yume: Nitpick: this being a bullet point feels a bit out-of-place.

More importantly though, you should say which libraries are used (SDL2, OpenAL) and mention the consequence of this (you need to install them from your package manage).

src/cpu/dynarmic_wrapper/build.rs
hikari-no-yume: I guess we could have the “static” feature control this, but I don't think it's worth the extra complexity for now. Would you be okay with setting `DYNARMIC_USE_BUNDLED_EXTERNALS` always and then removing the static feature for this package?

GeffDev: Should we be setting that flag for static builds?

hikari-no-yume: That would be the idea, though my preference would be to do it always and never attempt dynamic linking for the dynarmic dependencies, at least for now.

GeffDev:
Whoops, forgot to remove the static feature from the dynarmic crate.

src/cpu/dynarmic_wrapper/build.rs
```diff
@@ -72,16 +73,11 @@ fn main() {
...
     );
-    // fmtd isn't on Linux
-    if !cfg!(target_os = "linux") {
-        link_lib(if cfg!(debug_assertions) {
-            "fmtd"
-        } else {
-            "fmt"
-        });
+    link_lib(if cfg!(debug_assertions) {
+        "fmtd"
```
hikari-no-yume: Ah, is it named fmtd on Linux too then, when statically linked?

GeffDev: Seems like it. Compiles fine for me on debug and release with this.

hikari-no-yume:
Thanks for implementing my feedback! This looks good to me now. I'm going to test it out on my macOS and Windows setups and if everything looks good I'll merge it.

FYI: I basically never use traditional merges, I prefer to squash things into a single commit, since it makes for a cleaner git history. I'll do this when merging, but if you prefer, you can do the squash yourself first. (If you do, please make the commit author date reflect the time of the squash or the time the last commit was made, though.)

GeffDev:
Nice! Let me know if there are any issues with it on another platforms. 

I'm not too sure how to squash my commits, I'll let you do that haha.

hikari-no-yume:
Thanks, it's merged now as 73faa98e17cecb63f088a6f3a82cc9c263a35d65.

I made a follow-up commit (ad90aa71ab5d370cc2f3930a1138b6e46461b256) to update the changelog (please let me know if you'd prefer to be credited a different way) and also note that only OpenAL Soft works on macOS. I did actually try using the system OpenAL.framework, but apparently it has different behaviour, so I'm not going to recommend it.

GeffDev:
Thank you for merging!

GeffDev:
By the way, is there any chance I could get the dev discussions role in the discord server? I've been wanting to contribute to this project, namely looking at getting the emulated Sonic 1 port up and running. All good if not.

Issue #15, opened 2023-02-04
Poor quality text rendering: overlapping letters (particularly W and A) cut eachother off


hikari-no-yume:
You can notice this especially in some of the tutorial and instruction screens in Super Monkey Ball.

This happens because of how `cg_bitmap_context.rs`, `ui_font.rs` and `font.rs` interact. In short: there's no blending done, so transparent pixels (strictly speaking, texels) in one character's rectangle can obscure opaque pixels from an overlapping one.

A quick fix would be to add blending in `cg_bitmap_context.rs`, and that definitely should be done eventually, but I bet that falls down if you want to do stuff like rendering text with semi-transparency. Although to be fair, I've no idea if iPhone OS itself can do that correctly.

hikari-no-yume:
Done by @Xertes0 in https://github.com/hikari-no-yume/touchHLE/pull/91, merged as 9fc25fedb2345ab7d66b42d66e72fcb710582d31.

PR #91, opened 2023-03-24
Add color blending in `cg_bitmap_context`

Xertes0:
Fixes overlapping letters not displaying correctly - #15.

Should we also blend the alpha channel when `data.alpha_info == kCGImageAlphaOnly`?

### DIFF
--- a/src/frameworks/core_graphics/cg_bitmap_context.rs
+++ b/src/frameworks/core_graphics/cg_bitmap_context.rs
@@ -110,6 +110,92 @@ fn get_pixels<'a>(data: &CGBitmapContextData, mem: &'a mut Mem) -> &'a mut [u8] 
     mem.bytes_at_mut(data.data.cast(), pixel_data_size)
 }
 
+fn blend_alpha(bg: f32, fg: f32) -> f32 {
+    // Alpha is blended the same way in
+    // premultiplied and straight representation.
+    fg + bg * (1.0 - fg)
+}
+
+/// Blends two RGBA non gamma-encoded values, with straight alpha.
+fn blend_straight(bg: (f32, f32, f32, f32), fg: (f32, f32, f32, f32)) -> (f32, f32, f32, f32) {
+    if fg.3 == 0.0 {
+        // If fg.3 == 0.0 we attempt to blend fully transparent color.
+        bg
+    } else {
+        let new_a = blend_alpha(bg.3, fg.3); // Can't be 0 if fg.3 != 0
+        (
+            (fg.0 * fg.3 + bg.0 * bg.3 * (1.0 - fg.3)) / new_a,
+            (fg.1 * fg.3 + bg.1 * bg.3 * (1.0 - fg.3)) / new_a,
+            (fg.2 * fg.3 + bg.2 * bg.3 * (1.0 - fg.3)) / new_a,
+            new_a,
+        )
+    }
+}
+
+/// Blends two RGBA non gamma-encoded values, with premultiplied alpha.
+fn blend_premultiplied(bg: (f32, f32, f32, f32), fg: (f32, f32, f32, f32)) -> (f32, f32, f32, f32) {
+    (
+        fg.0 + bg.0 * (1.0 - fg.3),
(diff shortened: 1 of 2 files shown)


hikari-no-yume:
Hi, thanks for fixing this!

> Should we also blend the alpha channel when `data.alpha_info == kCGImageAlphaOnly`?

I think so. Alpha-only would be a natural format to use when rendering text to a texture. Red = green = blue = alpha when you're rendering white text on a transparent background.

Xertes0:
Okay, did that as well.

I also noticed that I was multiplying rbg by alpha twice so that's fixed.

hikari-no-yume:
Hmm, I guess I have three main comments on the code:

* Is there any particular reason you used integer rather than float arithmetic? There's nothing wrong with it in principle, it just doesn't match the surrounding code very well.
* This does alpha blending on gamma-encoded data, without any correction for this. Is that normal? I guess the code would become more complicated if it tried to do alpha blending in linear space.
* It might be nice to factor out the blending into a function, but maybe that's excessive.

hikari-no-yume:
FWIW I previously wondered about having a similar “get pixel” function. That would make doing alpha blending on linear RGB floats would be easy, but that'd be a lot of new lines of code.

Xertes0:
There was no particular reason to why I used integers, but I changed it to floats now. As you said it matches the surrounding code better.

I totally missed the gamma encoding. Now the blending is applied in linear space, and I did double check with [wikipedia](https://en.wikipedia.org/wiki/Alpha_compositing), so it should be correct this time.

I also factored out the blending functions.

As you mentioned it'd probably be better to add some kind of `get_pixel` and maybe `set_pixel` functions, that would shorten the match expressions. The question is what should `get_pixel` return when we only have the alpha channel in the bitmap? Setting the other channels to 1s might be confusing (or not?).

src/frameworks/core_graphics/cg_bitmap_context.rs
```diff
@@ -135,9 +167,50 @@ fn put_pixel(
...
+                pixels[first_component_idx + 3] as f32 / 255.0,
+            ),
+            (r, g, b, a),
+        ),
+        kCGImageAlphaFirst => blend_straight(
+            (
+                gamma_decode(pixels[first_component_idx + 3] as f32 / 255.0),
+                gamma_decode(pixels[first_component_idx] as f32 / 255.0),
+                gamma_decode(pixels[first_component_idx + 1] as f32 / 255.0),
+                pixels[first_component_idx + 2] as f32 / 255.0,
```
hikari-no-yume: Assuming you want to construct a tuple in (r, g, b, a) order, I think this is wrong? It doesn't match the code below that writes the new pixel out, at least.

Ditto for kCGImageAlphaPremultipliedFirst below.

hikari-no-yume:
Thanks for reworking this! The new code makes sense to me, aside from one comment.

> The question is what should `get_pixel` return when we only have the alpha channel in the bitmap? Setting the other channels to 1s might be confusing (or not?).

It doesn't really matter, but I think you could make the red, green and blue values be copies of the alpha channel.

src/frameworks/core_graphics/cg_bitmap_context.rs
Xertes0: Yeah, the order is incorrect, didn't notice.

Xertes0:
Added `get_pixel` function.

hikari-no-yume:
Thanks, the new version looks great! By the way, you might want to use `--reset-author` when doing a `git commit --amend` in future, so that the author date lines up with the latest changes, rather than preserving the old date. (I'm more particular about this than most people…) Do you mind if I fix that when cherry-picking?

Xertes0:
I don't mind.

hikari-no-yume:
Merged as 9fc25fedb2345ab7d66b42d66e72fcb710582d31. Thanks! I updated the changelog in b02f6f7a4d52202a1c8620fb710a43351c49554e, please let me know if you'd prefer to be credited in a different way.

PR #93, opened 2023-04-08
Add Github Action workflows for MacOS and Windows builds

dylanpdx:
This adds two Github Action jobs which build touchHLE and upload the results as an artifact. The uploaded artifact can be un-zipped and run just like a regular release, as I've included the `touchHLE_dylibs` and `touchHLE_fonts` directories.

### DIFF
--- a/.github/workflows/touchHLE_release.yml
+++ b/.github/workflows/touchHLE_release.yml
@@ -1,0 +1,65 @@
+name: Build touchHLE
+
+on:
+  push:
+    branches: [ "trunk" ]
+  pull_request:
+    branches: [ "trunk" ]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build-osx:
+
+    runs-on: macos-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Get Submodules
+      run: git submodule update --init
+    - name: Install Boost
+      run: brew install boost
+    - name: Build
+      run: cargo build --release && mv target/release/touchHLE .
+    - uses: actions/upload-artifact@v3
+      with:
+        name: touchHLE_macOS_x86_64
+        path: |
+          touchHLE_dylibs/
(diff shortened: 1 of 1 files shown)


hikari-no-yume:
Oh cool! I've never used GitHub's CI system before, I'm pleasantly surprised by how simple this script is.

Did you consider using the release prep script? I guess it might not work on Windows without git bash or something…

hikari-no-yume:
Hmm, maybe it's good that it doesn't include everything that a full release has, though…

dylanpdx:
> Did you consider using the release prep script? I guess it might not work on Windows without git bash or something…

I considered it, but wasn't sure if you would have preferred for me to differentiate it from a full release instead of a, 'pre-release' so to speak
I know the Windows runners do have Bash preinstalled so I think the release prep script *should* work, just a matter of installing pandoc and all that, but it shouldn't be an issue

hikari-no-yume:
Hmm… I could modify the release script so it has a mode where it does a bare-minimum release: fonts, dylibs, default options, and license info, but no readmes.

But if that's the goal… what if the build artifacts would be just the executables? It would save space. You can just grab the other files from git after all, or from a full release.

dylanpdx:
That could work! I hadn't considered just keeping the executables on their own, since originally my idea was for someone to be able to download an artifact, unzip it, and test the latest changes. Makes sense to save space though and since the extra files won't change, you could use the files from the last 'official' release. Should be a quick change if you prefer it like that!

hikari-no-yume:
It saves me from having to maintain more than one list of files to go in releases, so I think I'd like it.

If it ends up being annoying, it can be changed later I suppose.

hikari-no-yume:
Thanks, I'll play around with this later and then probably squash merge it. 👍

hikari-no-yume:
Forgot about this, but it's merged now :)

PR #94, opened 2023-04-09
Add GHA build

gliptak:
https://github.com/gliptak/touchHLE/actions/runs/4651782545

missed #93

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,49 @@
+name: ci
+
+on:
+  push:
+    branches:
+      - trunk
+  pull_request:
+
+jobs:
+  lint:
+    runs-on: macos-latest
+    steps:
+      - name: Install MacOS Boost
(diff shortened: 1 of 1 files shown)


hikari-no-yume:
Well, um, as you said…

> missed #93

This is a duplicate.

gliptak:
kinda. this PR implements lint and also targets to have a build matrix for supported OSs

hikari-no-yume:
Having linting would be nice to have for sure. I'm a bit confused by the build matrix though… it's not providing any benefit if there's only a single OS in it, and the build directions you've used will only work on that OS.

gliptak:
for the build matrix, we can add linux and windows in follow PR and just have the OS specific steps different (removing duplication as compared to #93 )

hikari-no-yume:
I'll consider it as a future improvement, I guess.

By the way, why do you install Boost twice?

gliptak:
the lint job is independent from the matrixed relerase_build jobs

PR #95, opened 2023-04-13
Added UIColor initWithWhite.

TylerJaacks:
According to my testing with Swift Playgrounds and reading this documentation on Apple's [website](https://developer.apple.com/documentation/uikit/uicolor/1621944-initwithwhite). initWithWhite takes in a White value and Alpha value and the is clamped between 0.0 and 1.0 on older versions of iOS, otherwise the rgb values take on the values of white and a takes on the value alpha.

After leaving a poor a commit I spent some time researching the code and "understand" how it works, this isn't a big change but it is something.


hikari-no-yume:
How have you tested this?

TylerJaacks:
I have not, I don't have a iPhone OS 2.0 SDK setup to create a app that uses this function.

src/frameworks/uikit/ui_color.rs
```diff
@@ -91,6 +91,18 @@ pub const CLASSES: ClassExports = objc_classes! {
 
 // TODO: more initializers, set methods, more accessors
 
+- (id)initWithWhite:(CGFloat) w alpha:(CGFloat) a {    
+    if w < 0.0 {
```
hikari-no-yume: Should there not be similar clamping for the alpha part too?

By the way, you can use Rust's `.clamp()` method on floats.

TylerJaacks: I don't if it should be clamped because in the documentation it only mentions the white arg being clamped "white values below 0.0 are interpreted as 0.0, and values above 1.0 are interpreted as 1.0."

hikari-no-yume: What documentation are you reading? [developer.apple.com says](https://developer.apple.com/documentation/uikit/uicolor/1621944-initwithwhite):

> Alpha values below 0.0 are interpreted as 0.0, and values above 1.0 are interpreted as 1.0.

TylerJaacks: I'm an idiot I will make this fix.

src/frameworks/uikit/ui_color.rs
```diff
@@ -91,6 +91,15 @@ pub const CLASSES: ClassExports = objc_classes! {
 
 // TODO: more initializers, set methods, more accessors
 
+- (id)initWithWhite:(CGFloat) w alpha:(CGFloat) a {        
+    let _w = w.clamp(0.0, 1.0);
+    let _a = a.clamp(0.0, 1.0);
```
hikari-no-yume: Thanks for the changes, it looks good now :)

I want to make a small suggestion though: in Rust, the `_` prefix is used for a value we don't care about and which won't be used (it silences “unused variable” warnings), but these variables are used. Perhaps you can do `let w` and `let a` instead of `let _w` and `let _a`. Rust lets you “shadow” the existing `w` and `a` variables, I use this a lot.

src/frameworks/uikit/ui_color.rs
hikari-no-yume, line 1: For symmetry, would you mind adding a `colorWithWhite:alpha:` method also? You can use the `colorWithRed:green:blue:alpha:` method as a reference.

TylerJaacks: yes

TylerJaacks:
I have implemented both changes.

src/frameworks/uikit/ui_color.rs
hikari-no-yume: Did you miss this comment?

src/frameworks/uikit/ui_color.rs
```diff
@@ -67,6 +67,13 @@ pub const CLASSES: ClassExports = objc_classes! {
     autorelease(env, new)
 }
 
++ (id)colorWithWhite:(CGFloat)w
+             alpha:(CGFloat)a {
```
hikari-no-yume: Nitpick: the style touchHLE tries to follow for Objective-C methods is that you align the colons on each line, so:

```objc
 (id)colorWithWhite:(CGFloat)w
              alpha:(CGFloat)a {
```

src/frameworks/uikit/ui_color.rs
```diff
@@ -91,6 +98,15 @@ pub const CLASSES: ClassExports = objc_classes! {
 
 // TODO: more initializers, set methods, more accessors
 
+- (id)initWithWhite:(CGFloat) w alpha:(CGFloat) a {
```
hikari-no-yume: Nitpick: can you remove the space between `(CGFloat)` and `w`, and `(CGFloat)` and `a`, to match other methods?

src/frameworks/uikit/ui_color.rs
hikari-no-yume: Though since there's only two arguments here, you could make this into a one-liner. Your choice.

TylerJaacks: resolved

src/frameworks/uikit/ui_color.rs
hikari-no-yume: It's fine if you don't want to change this, but it doesn't seem like you changed it…

hikari-no-yume:
It's merged now. Thanks!

PR #101, opened 2023-04-24
Add set osabi to debug instructions

abnormalmaps:


DEBUGGING.md
```diff
@@ -25,9 +25,10 @@ The basic set of steps is:
 * Start touchHLE in debugging mode: `touchHLE --gdb=localhost:9001 'Some App.app'`.
 * In a separate terminal window, start GDB: `gdb 'Some App.app/SomeApp'`. (You can omit the executable path, but this leaves GDB with no debug symbol info, [which is a worse experience](https://sourceware.org/bugzilla/show_bug.cgi?id=30234).) Then, inside GDB:
   * `set arch armv6`
+  *`set osabi none`
```
hikari-no-yume, line 28: The missing space here breaks the formatting.

hikari-no-yume:
Frustratingly, it doesn't seem like touchHLE can tell the gdb client which architecture and OS to use, at least not without sending a massive XML blob that I don't want to bundle.

DEBUGGING.md
hikari-no-yume: (If you want, I can fix this for you and commit it on your behalf.)

hikari-no-yume:
Oh, actually the XML is pretty simple, so I've implemented that now: f71690a5c2e0622810acdf697a4e798f6efa97ed. This makes your change redundant, sorry about that.

PR #104, opened 2023-04-27
Add basic objc_sync_enter/exit (@synchronized) support

abnormalmaps:
This adds support for objc_sync_enter and objc_sync_exit, which are used to implement @synchronized blocks. I've only tested it  with one app (DDRSEU.app 1.1.3, has some other stub-ish patches). I'd like to see if I can get proper thread blocking for mutexes and @synchronized at some point, so I'll add tests then.


abnormalmaps:
(Apologies to `@synchronized` for the notification, oops)

hikari-no-yume:
Ooh, interesting! Most of my knowledge of Objective-C comes from this project, and I hadn't encountered this feature yet, hehe.

hikari-no-yume:
Apple's comments suggest they use pthread mutexes. Any particular reason you aren't using touchHLE's pthread implementation? There's no actual waiting for unlock implemented yet, but when that is eventually implemented, it would be nice to share it.

abnormalmaps:
> Apple's comments suggest they use pthread mutexes. Any particular reason you aren't using touchHLE's pthread implementation?

I didn't really want to deal with the round trip of allocating into guest memory for it to be used only by the host. That said, it might be better to have an internal pthread-like interface to creating/locking/unlocking mutexes, because it'll probably also be needed for other locks (mainly the NSLock family).

I'll close this for now since I'd prefer to get it fully working with thread blocking at this point.

PR #112, opened 2023-05-11
Improve support for some apps built for IOS 3

abnormalmaps:
This adds functionality that seem to be used in some apps that are built for IOS 3. (specifically, I'm building for IOS 3 on XCode 4.4, because it's the earliest version that I can run) The main two things added are:

- NSStringFromClass, which seems to be used in calling UIApplicationMain
- UIApplicationMain checking and prioritizing application:didFinishLaunchingWithOptions:, which seems to be the preferred for IOS >3. This isn't strictly necessary, but

Other than putchar (which i added since it doesn't justify it's own PR), most of the other functions are added to make the main two things work.


src/frameworks/uikit/ui_application.rs
```diff
@@ -182,15 +182,19 @@ pub(super) fn UIApplicationMain(
         let delegate: id = msg![env; ui_application delegate];
         // IOS 3+ apps usually use application:didFinishLaunchingWithOptions:, and it
         // seems to be prioritized over applicationDidFinishLaunching:.
-        if let Some(launch_with_options) = env.objc.lookup_selector("application:didFinishLaunchingWithOptions:") {
-            if msg![env; delegate respondToSelector:launch_with_options] {
+        match env
+            .objc
+            .lookup_selector("application:didFinishLaunchingWithOptions:")
+        {
+            Some(launch_with_options) if msg![env; delegate respondsToSelector:launch_with_options] =>
```
hikari-no-yume: There's going to be many more optional methods like this. Could you extract this into a helper function? The signature would be something like:

```rust
fn responds_to_selector(env: &mut Environment, selector: &str) -> bool;
```

It could go in `src/objc/objects.rs` perhaps. I'm suggesting a function rather than a method because it'd need to take a reference to `env` and is quite high-level. Please include a documentation comment about what it does and why (which would be obvious to you I guess).

src/objc/classes.rs
```diff
@@ -675,3 +676,10 @@ impl ObjC {
         }
     }
 }
+
+pub fn class_getName(env: &mut Environment, class: id) -> ConstPtr<u8> {
+    let class_ptr: ConstPtr<class_t> = ConstPtr::from_bits(class.to_bits());
+    let class_data: class_t = env.mem.read(class_ptr);
+    let class_name: ConstPtr<u8> = env.mem.read(class_data.data).name;
+    class_name
+}
```
hikari-no-yume, line 710: Unfortunately, this won't always work. For classes defined by touchHLE itself (i.e. all the classes in our framework implementations), the class allocation isn't `class_t` and its name only exists as a Rust string inside its host object. That could be changed, I suppose, but I think I'd prefer allocating such strings on-demand. I'm assuming that the exact shape of `class_t` is a detail app code isn't expected to rely on.

hikari-no-yume (changes requested):

hikari-no-yume:
Thanks for your changes. It's good to see that iPhone OS 3 apps work at some basic level… I thought there'd be more challenges, but maybe I was looking at an app for 3.1 or 3.2 or something, which is a bit more complex.

src/objc/classes.rs
hikari-no-yume: With that in mind, by the way, `NSStringFromClass` could potentially be written so that it creates the `NSString` from the Rust string directly. I don't have an opinion on whether it should, though I will say it's more efficient that way.

src/frameworks/uikit/ui_application.rs
abnormalmaps: Split this out, also added instance_responds_to_selector. Didn't use the same implementation as in NSObject because I'd like not to rely on objects subclassing NSObject.

src/objc/classes.rs
abnormalmaps: Didn't realize this at the time ._. . Should hopefully be fine now, it seems to have worked with my quick test at least.

abnormalmaps:
> It's good to see that iPhone OS 3 apps work at some basic level… I thought there'd be more challenges, but maybe I was looking at an app for 3.1 or 3.2 or something, which is a bit more complex.

Admittedly, I've only tested this on the most barebones app possible (more or less the "Empty Application" sample), since this is mostly so I can build tests for myself. On that note, I can write a test for NSStringFromClass, but it'll probably come in a later PR since i forgot the first time and don't want to rerun the build.

src/frameworks/foundation/ns_util.rs
```diff
@@ -0,0 +1,24 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+//! Functions under `Objective-C Runtime Utilities`, just `NSStringFromClass` right now.
+
+use crate::{
```
hikari-no-yume: Stylistic nit: For consistency with the rest of the code, could you use the pattern `use crate::module::item;`/`use crate::module::{item1, item2};`? There's nothing really wrong with this style, it's just quite different from all the rest of the code.

Also, uh, please import `export_c_func` from `dyld` rather than the crate root. It's only exported from the latter due to scoping quirks.

src/libc/stdio.rs
```diff
@@ -10,6 +10,7 @@ use crate::dyld::{export_c_func, FunctionExports};
 use crate::fs::GuestPath;
 use crate::mem::{ConstPtr, ConstVoidPtr, GuestUSize, MutPtr, MutVoidPtr, Ptr, SafeRead};
 use crate::Environment;
+use core::slice;
```
hikari-no-yume: Stylistic nit: Could you import this from `std` instead? It doesn't matter of course. Also FWIW, for the freestanding functions in `std::mem`, `std::str` and `std::slice`, I usually just write out their full name at the callsite rather than importing them.

src/objc/classes.rs
```diff
@@ -16,6 +16,7 @@ pub(super) use class_lists::CLASS_LISTS;
 use super::{
     id, method_list_t, nil, objc_object, AnyHostObject, HostIMP, HostObject, ObjC, IMP, SEL,
 };
+use crate::environment::Environment;
```
hikari-no-yume: Stylistic nit: For historical reasons, `Environment` should be imported directly from the crate root.

src/objc/objects.rs
```diff
@@ -259,3 +260,30 @@ impl super::ObjC {
...
     }
 }
+
+/// Checks if an instance of the provided class responds to a selector (has a method in it's class
+/// chain) with the given name.
+///
+/// Note that this only checks instances of the class, not the class itself.
+/// This means that class methods of the provided class are not checked, only instance methods.
+pub fn instances_respond_to_selector(
+    env: &mut Environment,
```
hikari-no-yume: Hmm… I think this gets quite complicated when we try to use the same naming that Objective-C does, and it's even a bit misleading since it doesn't take a `SEL` like its Objective-C counterpart. What if `class_has_method` had a sibling method in the same file, named something like `class_has_method_named`? That would be a better fit for the current organisation of things. The description would only need to say that it does the same thing as `class_has_method`, except it doesn't panic or do an allocation when the selector is unknown. I guess `object_has_method` could be the variant that does the `read_isa` for you.

Also FWIW you could use the new `let … else` syntax in this method body to simplify it slightly, if you wanted.

src/objc/classes.rs
```diff
@@ -675,3 +676,29 @@ impl ObjC {
...
+
+pub fn class_getName(env: &mut Environment, class: Class) -> ConstPtr<u8> {
+    let class_name = class_getName_inner(&mut env.objc, class);
+    env.mem
+        .alloc_and_write_cstr(class_name.as_bytes())
+        .cast_const()
+}
+
+/// Internal method that gets class name as a &str.
+pub fn class_getName_inner(objc: &mut ObjC, class: Class) -> &str {
```
hikari-no-yume: This inner function should probably be a method on `ObjC`. It could be called `get_class_name` I suppose.

hikari-no-yume (changes requested):
First of all, I'm sorry it's taken me so long until I looked at this again. I really appreciate your efforts!

I have some further suggestions. They're a bit nitpicky, sorry; if any of it feels too much, I'm happy to make the changes myself.

When merging this I'll probably rewrite the history so each change is self-contained (no fixup commits), then cherry-pick those rewritten changes. If you'd prefer to do this yourself, let me know.

abnormalmaps:
Thanks for getting around to it! Please don't feel any pressure to merge or ask for changes - I don't want me or the project to be a burden to you. 

I'll make the clean up commits, but I'd like you to squash the commits since I don't know exactly how you want them done. Thank you!

src/frameworks/foundation/ns_util.rs
abnormalmaps: I'm going to be honest, I pretty much just used my LSP auto-import for pretty much all of the imports. Most of the other ones I'll need to clean up manually, but for this one there's actually a rustfmt lint, so I'll add it in and you can merge if you want.

hikari-no-yume: Ah I see. If there's some way I can help LSPs to match what I do manually (or vice-versa), let me know.

abnormalmaps: Just realized it's unstable :P. Probably won't add it then.

hikari-no-yume: All the good lints are :(

abnormalmaps:
All of the fixes should be in. There's also a small fix for the Cargo.toml: it used to build SDL even if `--no-default-features` was on, which was a bit annoying and seemed to be duplicated from the default features. If you don't want to merge that one it's OK.

hikari-no-yume:
Oh I see, so that's what the Cargo.toml change was about. I think the duplication is accidental and happened when I was merging the Android support (b7792fa77351da821b7d2e9bde398de7c00af39b). Thanks for catching it. :)

hikari-no-yume:
I cherry-picked/squashed a few things to trunk now:

* 1881e89974d59213f2ce075a90b569889e94d957 (the SDL2 thing)
* ad17bd572cd89e76359a5973a5928cfa2987efe3 (the selectors thing)
* ed0226bbe90dab38f1702a7cdad0793ac0691058 (`putchar`)
* 3746d606db81c9b4c2b5d3d8e25a6fd2b04482bc (the `NSDictionary` methods)

Thanks for your efforts!

I'm holding off on the class name stuff for the moment because I noticed some issues.

hikari-no-yume:
I implemented some of the class name stuff myself in 0f15a4ecf087b88fd7212c781a620aa352dd980f because I wasn't quite happy with the version in your PR or Lywx's PR. I haven't done `class_getName` so far but based on what you said, you didn't actually see it used in the apps you were interested in. So I think I can close this now.

hikari-no-yume:
Oh, I should say, I'm sorry I ended up wasting your efforts there. It's just a case where it seemed like it'd save time and effort for both of us if I just redo it the way I want it myself.

PR #114, opened 2023-05-13
Support for Wolf3D-iOS app

ciciplusplus:
This PR introduces support for Wolfenstein 3D for iOS.

### Compatibility

Version 1.0 (versions 1.1+ will crash on start ~~and even if crash is fixed there are texture decoding problems~~)

First level is definitely playable :)

### Known issues

- Freeze on audio loading (first loading of an audio/sound will freeze touchHLE. I suspected inefficient realloc implementation and even attempted an improved version [here](https://github.com/ciciplusplus/touchHLE/commit/046f4f6d745b92497407303e9fd6d42562331aff), but it seems not to solve issue fully)
- Multi-touch is not implemented

### Screenshots

Screenshot 2023-05-13 at 13 41 30
Screenshot 2023-05-13 at 13 41 08
Screenshot 2023-05-13 at 13 41 00

### Video

https://twitter.com/ciciplusplus/status/1656388212405805065


src/frameworks/audio_toolbox/audio_services.rs
```diff
@@ -33,4 +35,13 @@ fn AudioServicesGetProperty(
     }
 }
 
-pub const FUNCTIONS: FunctionExports = &[export_c_func!(AudioServicesGetProperty(_, _, _, _, _))];
+fn AudioServicesPlaySystemSound(_env: &mut Environment, in_system_sound_id: SystemSoundID) {
+    assert_eq!(in_system_sound_id, kSystemSoundID_Vibrate)
+    // TODO: vibrate on devices supporting it
```
hikari-no-yume: Could you make this TODO into a `log!` call, something like `log!("TODO: vibration (AudioServicesPlaySystemSound)");`? That way the user at least knows the game _should_ have vibrated.

src/frameworks/core_graphics/cg_image.rs
```diff
@@ -120,11 +121,17 @@ fn CGImageGetHeight(env: &mut Environment, image: CGImageRef) -> GuestUSize {
     height
 }
 
+fn CGImageGetDataProvider(_env: &mut Environment, image: CGImageRef) -> CGDataProviderRef {
+    // TODO: implement proper provider?
```
hikari-no-yume: Can you add a comment mentioning the hack in cf_data.rs and cg_data_provider.rs?

src/frameworks/foundation/ns_data.rs
```diff
@@ -7,6 +7,7 @@
 
 use super::ns_string::to_rust_string;
 use super::NSUInteger;
+use crate::environment::Environment;
```
hikari-no-yume: Please import `Environment` directly from the crate root (maybe I should change this convention but it's for historical reasons).

src/frameworks/opengles/gles1_native.rs
```diff
@@ -545,4 +548,7 @@ impl GLES for GLES1Native {
     unsafe fn DeleteRenderbuffersOES(&mut self, n: GLsizei, renderbuffers: *const GLuint) {
         gles11::DeleteRenderbuffersOES(n, renderbuffers)
     }
+    unsafe fn GenerateMipmapOES(&mut self, target: GLenum) {
```
hikari-no-yume: Could you move this to the textures section? (Also in all the other files.)

src/frameworks/uikit/ui_image.rs
```diff
@@ -7,6 +7,7 @@
 
 use crate::frameworks::core_graphics::cg_image::{self, CGImageRef, CGImageRelease};
 use crate::frameworks::core_graphics::CGSize;
+use crate::frameworks::foundation::ns_data::to_rust_slice;
```
hikari-no-yume: Can you change the import to be just `ns_data` (you can merge it with the following line), and then write `ns_data::to_rust_slice` at the callsite? That's what I do for these pseudo-methods on `ns_string`.

src/frameworks/uikit/ui_touch.rs
```diff
@@ -16,6 +17,29 @@ use crate::objc::{
...
+#[repr(i32)] // this should be an NSInteger!
+pub enum UITouchPhase {
+    UITouchPhaseBegan = 0,
+    UITouchPhaseMoved = 1,
+    UITouchPhaseEnded = 3,
+}
+
+impl GuestRet for UITouchPhase {
+    fn from_regs(regs: &[u32]) -> Self {
+        // TODO: how to simplify?
```
hikari-no-yume: Rust doesn't provide a _safe_ simpler way alas :/

src/frameworks/uikit/ui_touch.rs
```diff
@@ -16,6 +17,29 @@ use crate::objc::{
 use crate::window::Event;
 use crate::Environment;
 
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+#[repr(i32)] // this should be an NSInteger!
+pub enum UITouchPhase {
+    UITouchPhaseBegan = 0,
+    UITouchPhaseMoved = 1,
+    UITouchPhaseEnded = 3,
+}
```
hikari-no-yume: For these C-like enums I do this instead (you can see this pattern in many files of touchHLE):

```rust
pub type UITouchPhase = NSInteger;
pub const UITouchPhaseBegan: UITouchPhase = 0;
// etc
```

It's a deviation from typical Rust but it avoids needless conversion back and forth, and acknowledges that C enums aren't really type-safe anyway. I should probably make a macro for this pattern so it's less repetitive.

src/libc/dlfcn.rs
```diff
@@ -12,18 +12,23 @@ use crate::Environment;
 fn dlopen(env: &mut Environment, path: ConstPtr<u8>, _mode: i32) -> MutVoidPtr {
     // TODO: dlopen() support for real dynamic libraries, and support for all
     // libraries with host implementations.
-    assert_eq!(env.mem.cstr_at_utf8(path), Ok("/usr/lib/libSystem.B.dylib"));
+    assert!([
+        Ok("/usr/lib/libSystem.B.dylib"),
+        Ok("/System/Library/Frameworks/OpenAL.framework/OpenAL")
```
hikari-no-yume: Since we are repeating this same list in three places, maybe there should be a `const` for it above? Not important though.

src/libc/errno.rs
```diff
@@ -5,6 +5,37 @@
...
+        // TODO: "real" errno implementation
+        *self
+            .errnos
+            .entry(thread)
+            .or_insert_with(|| mem.alloc_and_write(0i32))
+    }
+}
+
+fn __error(env: &mut Environment) -> MutPtr<i32> {
+    env.libc_state
```
hikari-no-yume, line 38: Can you add something like `log!("TODO: errno (returned 0)");`? If we're always returning `0`, I'm sure _something_ will break, and it'd be good to at least have a warning about it in the log.

src/libc/posix_io/stat.rs
```diff
@@ -41,7 +41,7 @@ fn fstat(env: &mut Environment, fd: FileDescriptor, buf: MutVoidPtr) -> i32 {
     // TODO: error handling for unknown fd?
     let file = env.libc_state.posix_io.file_for_fd(fd).unwrap();
 
-    log!("Warning: fstat() call, this function is mostly unimplemented");
+    log_dbg!("Warning: fstat() call, this function is mostly unimplemented");
```
hikari-no-yume: How often does Wolf3D call fstat()? :p

src/libc/stdio/printf.rs
```diff
@@ -41,12 +41,16 @@ pub fn printf_inner<const NS_LOG: bool, F: Fn(&Mem, GuestUSize) -> u8>(
...
+        let mut pad_char = if get_format_char(&env.mem, format_char_idx) == b'0' {
             format_char_idx += 1;
             '0'
         } else {
             ' '
         };
+        if get_format_char(&env.mem, format_char_idx) == b'.' {
+            format_char_idx += 1;
+            pad_char = '0';
+        }
```
hikari-no-yume, line 57: Huh, this confused me, but it seems like this is more or less accurate for integers… but for floats, `.` does something different. Can you add some kind of assert?

src/libc/stdio/printf.rs
```diff
@@ -209,7 +217,62 @@ fn printf(env: &mut Environment, format: ConstPtr<u8>, args: DotDotDot) -> i32 {
...
+
+        let specifier = env.mem.read(format + format_char_idx);
+        format_char_idx += 1;
+
+        match specifier {
+            b'd' => {
+                let mut val: i32 = 0;
+                while let c @ b'0'..=b'9' = env.mem.read(src_ptr) {
+                    val = val * 10 + (c - b'0') as i32;
```
hikari-no-yume, line 266: 😍

src/objc/classes.rs
```diff
@@ -395,11 +395,11 @@ fn substitute_classes(
     let name = mem.cstr_at_utf8(name).unwrap();
 
     // Currently the only thing we try to substitute: classes that seem to be
-    // from the AdMob SDK. This is a third-party advertising SDK. Naturally it
+    // from the AdMob SDK (or other ads SDK). This is a third-party advertising SDK. Naturally it
```
hikari-no-yume: Maybe just say “from various third-party ad SDKs”.

hikari-no-yume (changes requested):
Nice work! Looks a lot less hacky than the prototype did ;) Sorry I took so long to get round to reviewing it…

I have a few suggestions but I think I'm generally happy with this.

hikari-no-yume:
Oh right: if you can make changes by amending git history rather than adding new commits, I would appreciate it. I can do this for you if necessary though.

src/libc/posix_io/stat.rs
ciciplusplus: I have about 186 entries of this log on the startup of the game. Basically each load of a new resource will produce this log as game checks size first.

But I can change back, as you wish :p

src/frameworks/uikit/ui_touch.rs
ciciplusplus: 😭

src/libc/posix_io/stat.rs
hikari-no-yume: Hmm… I wish it were possible to somehow `assert` that `fstat()` is only used to read the filesize :(

Here's an idea: you could add a state variable for “have warned about fstat” so the log message is only shown once? I do something like that in `time.rs`.

ciciplusplus: on the second thought, ~200 is not that much of logs, so i think i'll put log!() back

hikari-no-yume: Nobody actually reads them other than us devs anyway :p

src/libc/errno.rs
ciciplusplus: I can, but the problem is that it become very spammy - as Wolf3d loads a new chunk of ogg on each frame, it also check errno value :/

I can use log_dbg!(), is it acceptable ?

ciciplusplus: I checked that was done for time.rs, but I'm not a big fan of introducing a local state just to make logging better

src/frameworks/audio_toolbox/audio_services.rs
ciciplusplus: Done

src/frameworks/core_graphics/cg_image.rs
ciciplusplus: Done

src/frameworks/foundation/ns_data.rs
ciciplusplus: Done

src/frameworks/opengles/gles1_native.rs
ciciplusplus: Done

src/frameworks/uikit/ui_image.rs
ciciplusplus: Done

src/frameworks/uikit/ui_touch.rs
ciciplusplus: Done

It's actually simplified the code, I don't need to define GuestArgs anymore :)

src/libc/dlfcn.rs
ciciplusplus: Done

src/libc/stdio/printf.rs
ciciplusplus: Done

src/objc/classes.rs
ciciplusplus: Done

ciciplusplus:
> Oh right: if you can make changes by amending git history rather than adding new commits, I would appreciate it. I can do this for you if necessary though.

Done. I also rebased on the trunk, so it should be all good now 😸

src/frameworks/uikit/ui_touch.rs
hikari-no-yume: I think this even reduces code size bloat because it just reuses the `NSInteger` implementation of `GuestArgs` :D

src/libc/errno.rs
hikari-no-yume: I don't really want to use `log_dbg!` for this kind of thing, because you would only enable it if you already suspected `errno` is the problem.

But I just realised, we don't need any new state! You are already creating the errno variable on demand, you can put the `log!` in there (`log!("TODO: errno accessed on thread {} (will always be 0)");`)

hikari-no-yume:
Sorry, I didn't realise until now that glGenerateMipmap comes from the EXT/OES_framebuffer_object extensions, even though the suffix was staring me in the face, so I shouldn't have told you to move it. To save you further effort, I'll just cherry-pick your older commit (065cd03b1779048c617cc0833b64e0799b3b861b).

hikari-no-yume:
Actually, I don't like that `glGetString` leaks memory if you call it multiple times for the same string (not sure why an app would do that, but it could). I have an idea for how to cleanly avoid this so I think I'll just implement it myself.

hikari-no-yume:
I just cherry-picked a bunch of these to trunk:

03df4ad199574c2b57978588b3041dc85e2c4ea9 Game specifc audio hack: ignoring 0 doppler velocity
045c6517fb8b0783ad7101c317bd3c1da02782bd Fix crash on invoking vibration
d1a89df0ff2727c692f229bfb86bd17922634a3a Implement UITouchPhase and keep track of phase across touches
270d064443d0f23e9fc4e49bcb223cd9b32073b7 Implement bunch of CF/CG method to load and copy pixels data from an image
bb6f7bce154ffe9d3348a08628c57e60d8ca7c74 UIImage imageWithData
3bc7660552f4dac8f0fa27d2359b9ac46298e431 Ignore advertising SDKs (they are not loading ads anyway)
72323528e8711be0f7bbdfc138acb3fdb2cfefca Various UIKit related stubs
f2f5fee45212f117ec6ad88dcf6a68bf12a068dd .precision for integer specifiers in *printf*
40869503b26f058515dc31b433e849c74b1a1ab0 OpenAL stubs
bf9db3235bba214d547bd04db314c13f9bd2a05e %x specifier for *printf*
51a0cb848e23c5631552148e8aca98645e922000 sscanf() implementation + test

Because I skipped the errno one, I also had to skip the other ones that touch the tests, for the moment.

src/libc/time.rs
```diff
@@ -271,10 +271,8 @@ unsafe impl SafeRead for timeval {}
 fn gettimeofday(
     env: &mut Environment,
     timeval_ptr: MutPtr<timeval>,
-    timezone_ptr: MutVoidPtr, // deprecated, always NULL
+    _timezone_ptr: MutVoidPtr, // deprecated, not used
 ) -> i32 {
-    assert!(timezone_ptr.is_null());
```
hikari-no-yume, line 276: This is too hacky to me :p

Can you instead check what the actual timezone struct format is for iPhone OS, and return one that's all zeroes?

hikari-no-yume (changes requested):

src/libc/time.rs
hikari-no-yume: Looks like it is:

```c
struct timezone {
    int     tz_minuteswest; /* minutes west of Greenwich */
    int     tz_dsttime;     /* type of dst correction */
};
```

so it can be:

```rust
#[allow(non_camel_case_types)]
#[repr(C, packed)]
struct timezone {
    tz_minuteswest: i32,
    tz_dsttime: i32,
}
```

and then just set to zero (it just so happens that `tz_dsttime == 0` means UTC :)

src/libc/errno.rs
ciciplusplus: Right, done

src/libc/time.rs
ciciplusplus: Done

src/libc/time.rs
```diff
@@ -268,13 +268,19 @@ struct timeval {
...
 fn gettimeofday(
     env: &mut Environment,
     timeval_ptr: MutPtr<timeval>,
-    timezone_ptr: MutVoidPtr, // deprecated, always NULL
+    timezone_ptr: MutPtr<timezone>,
 ) -> i32 {
-    assert!(timezone_ptr.is_null());
-
     if timeval_ptr.is_null() {
         return 0; // success
```
hikari-no-yume, line 295: Oh, this needs to be changed now (don't skip setting the timezone_ptr if the timeval_ptr is null)

hikari-no-yume (changes requested):

src/libc/time.rs
ciciplusplus: Done

src/libc/stdio.rs
```diff
@@ -87,6 +88,13 @@ fn fgetc(env: &mut Environment, file_ptr: MutPtr<FILE>) -> i32 {
     }
 }
 
+fn fputs(env: &mut Environment, str: ConstPtr<u8>, stream: MutPtr<FILE>) -> i32 {
+    let str_len = strlen(env, str);
+    fwrite(env, str.cast(), str_len, 1, stream)
```
hikari-no-yume, line 94: This actually has the wrong behaviour on error (doesn't set errno or return EOF), maybe add a TODO comment?

hikari-no-yume (changes requested):

hikari-no-yume:
Cherry-picked four more commits to trunk:

4cf7902ced0f0e29c741c145370ef36a6cfeac90 gettimeofday can be actually called with non null timezone_ptr
35991d82ad2083e8510295a1e634f56f38ab3eb3 Handle mismatch in sscanf, more tests
a57a15314d8d4de15f8ae9de9005cc3bc619132f realloc implementation + test
91edf001a55b03239e57c90fecab6b42a43ad43e errno implementation + test

hikari-no-yume:
note to self, only remaining items for me to do are:

- [x] cherry-pick fputs once ciciplusplus is done with it
- [ ] ~~reimplement glGetString and glGenerateMipMapOES~~
- [x] fix NSFileManager thingie
- [x] add wolf3d to supported apps list

src/libc/stdio.rs
ciciplusplus: Done :)

hikari-no-yume:
Actually, reimplementing glGetString is pretty annoying and I cba right now, so I just added a warning instead in 521b3c157ee81af4e31736ed7b4f8b650daf6816 ;)

And with that… all the Wolf3D changes are on trunk!

hikari-no-yume:
```
$ git log -1
commit 521b3c157ee81af4e31736ed7b4f8b650daf6816 (HEAD -> trunk, hikari/trunk)
Author: hikari_no_yume <hikari@noyu.me>
Date:   Thu Jul 20 08:44:46 2023 +0200

    Add warning to glGetString()
```
Screenshot of Wolfenstein 3D running in touchHLE v0.1.2-217-g521b3c1. BJ is being shot at by two guards and is quite bloodied.

hikari-no-yume:
8bddf9199f1516a0b1d7ad7f64bdb71e72983c39 :)

Issue #119, opened 2023-05-29
Semi-co-operative threading model is prone to deadlocks when host-to-guest calls stack


hikari-no-yume:
touchHLE has a hybrid of co-operative and pre-emptive multitasking. There is full pre-emption for guest code (we just swap out the register state for threads like any OS would), but because everything runs on a single host thread and we don't use coroutines, host code is co-operative: other host or guest code can't execute until host code is done.

A simple case where this can be problem is with [the current implementation of `sleep()`](https://github.com/hikari-no-yume/touchHLE/commit/b6fe42693f56e8cab13ad27c9ecd69f9c1960905). If we have two threads which are both sleeping, the host call stack can become this mess:

* thread 1 calls `sleep()`
  * `Environment::run_call()` -> `Environment::run_inner()`
    * switches to thread 2
      * thread 2 calls `sleep()`
        * `Environment::run_call()` -> `Environment::run_inner()`
          * spins until thread 2 wakes up. **if thread 1 wakes up earlier, we are still stuck here!**
          * thread 2 wakes up and signals return-to-host
        * `sleep()` returns
    * waits until thread 1 wakes up
    * thread 1 wakes up and signals return-to-host
  * `sleep()` returns

Sleeping for longer than necessary is relatively benign, but there can be far worse consequences. Potentially we can end up with deadlocks.

In any case where

* guest code on thread A calls a host function,
* that function calls another guest function on thread A,
* while in that call, `Environment::run_inner()` switches threads to thread B,
* guest code on thread B calls a host function on thread B, and
* that function calls another guest function on thread B

then we will end up in a deadlock if the inner guest function on thread B is waiting on something in the outer guest function on thread A. This is because thread A's host function call can't return while we're busy with thread B's host function call.

These deadlocks could happen randomly! There's no guarantee of when thread switching will or won't happen. And of course, we have this guest-to-host-to-guest call structure very frequently, for example because of `objc_msgSend`.

This isn't a theoretical issue, @ciciplusplus has already hit problems related to this when trying to get DOOM working.

----

This is one of touchHLE's most fundamental design limitations, and it's not going to be easy to fix. There's only two solutions I can see:

[...]

PR #123, opened 2023-06-07
Implement thread blocking system & various blocking functions

abnormalmaps:
This PR adds a more generalized system for thread blocking, which is used to implement mutexes, `@synchronized`, `pthread_join`, and return-to-host deferral. In order:

- I've changed mutexes to use an internal id to index get the host object instead of the pointer to the mutex. This is so host objects can hold mutexes (right now just objc_sync_enter/exit, in the future the NSLock family will need it).
- objc_sync_enter.exit (`@synchronized`) is pretty much the same as in #104, just with mutexes instead of the bespoke system in that PR.
- The return-to-host deferral fixes a bug where secondary threads that aren't the top of the current call stack mistake a return-to-host for the thread ending.

I've done some basic testing, but I haven't made proper tests yet. Also this doesn't at all solve #119 (although I do have an idea based on this commit involving async, and if you're open to it I can give it a shot).


hikari-no-yume:
Thanks for your efforts! It might be a while before I can take a close look at this, but my initial impressions are good. The deferred return-to-host and improved `sleep()` are both things I was going to have to implement soon, and it seems like your implementations might be fairly clean, so you've probably saved me some effort.

> Also this doesn't at all solve https://github.com/hikari-no-yume/touchHLE/issues/119 (although I do have an idea based on this commit involving async, and if you're open to it I can give it a shot).

I was also wondering about using `async fn` for that. I've never used that feature, but from what little I know of it, it does seem like it would be the ideal solution here, so please do try it out if you feel like it. I anticipate the all-important `&mut Environment` could be a problem, since the function can't maintain a reference to it while suspended, and Rust doesn't seem to let us directly pass it in when resuming, so we might need to be able to pass it indirectly, or avoid using it somehow.

hikari-no-yume:
Oh, one more thing: if you want to access the secret developer discussions room of the Discord server, please send me an email (see the git history). Thanks!

src/environment.rs
```diff
@@ -573,49 +686,94 @@ impl Environment {
...
+                            }
+                        }
+                        ThreadBlock::Joining(joinee_thread, ptr) => {
+                            if !self.threads[joinee_thread].active {
+                                log_dbg!(
+                                    "Thread {} joining with now finished thread {}.",
+                                    self.current_thread,
+                                    joinee_thread
+                                );
+                                // Write the return valiue, unless the pointer to write to is null.
```
hikari-no-yume: s/valiue/value/

src/libc/pthread/mutex.rs
```diff
@@ -6,20 +6,28 @@
...
 use std::collections::HashMap;
 use std::num::NonZeroU32;
 
 #[derive(Default)]
 pub struct State {
-    mutexes: HashMap<MutPtr<pthread_mutex_t>, MutexHostObject>,
+    // TODO?: Maybe this should be a Vec instead? It would be bad if there were many mutexes over
+    // the lifetime of an application, but it would perform better.
+    // Maybe it could also be a fixed size allocator? (although that seems a little overkill)
+    mutexes: HashMap<HostMutexId, MutexHostObject>,
```
hikari-no-yume: Hmm, now that mutexes are not just a pthread thing, but something shared between various parts of touchHLE, I think the core parts shouldn't live here any more. I suggest making a new module for all the host/shared stuff. I guess the best place to put it would be `src/environment/mutex.rs`. I guess it could use a similar pattern to `ObjC`, using an `impl` block to add methods to `Environment`? (That type is pretty messy but I can't think of a better way for now…) Then this file can be purely the `pthread_` functions that are wrappers over it.

hikari-no-yume (changes requested):
Ah, this looks great! Sorry that I took so long before looking at it properly. I've probably missed some details, but overall I really like it. Aside from one organisational change, I'm pretty much happy to merge this as-is. If you make that change, I'll make a second pass to look at some style things, but I don't think I'll have much to comment in.

src/environment.rs
```diff
@@ -354,32 +391,68 @@ impl Environment {
...
-        let old_pc = self.cpu.pc_with_thumb_bit();
-        self.cpu.branch(self.dyld.return_to_host_routine());
-        // Since the current thread is asleep, this will only run other threads
-        // until it wakes up, at which point it signals return-to-host and
-        // control is returned to this function.
-        self.run_call();
-        self.cpu.branch(old_pc);
+        self.threads[self.current_thread].blocked_by = ThreadBlock::Sleeping(until);
+        // For non tail-call sleeps (such as in NSRunLoop), we want to poll other threads but can't
+        // return until the sleep in completed, so we'll run in a seperate
```
hikari-no-yume: This sentence looks unfinished.

src/environment.rs
```diff
@@ -354,32 +391,68 @@ impl Environment {
...
         self.switch_thread(old_thread);
 
         new_thread_id
     }
 
     /// Put the current thread to sleep for some duration, running other threads
-    /// in the meantime as appropriate.
-    pub fn sleep(&mut self, duration: Duration) {
-        assert!(self.threads[self.current_thread].sleeping_until.is_none());
+    /// in the meantime as appropriate. Functions that call sleep right before they return should set
```
hikari-no-yume: This should be qualified: before returning to the guest/`Environment::run`. Functions that can be called directly by other host functions shouldn't use this.

src/objc/synchronization.rs
```diff
@@ -0,0 +1,106 @@
...
+//! `@synchronized` blocks are sections of code that, for a given object, only allow one thread inside any `@synchronized` block with that object.
+//! These are internally implemented with the `objc_sync_enter` and `objc_sync_exit` functions.
+//!
+//! Resources:
+//! - [Section about `@synchronized` in *The Objective-C Programming
+//! Language*](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjectiveC/Chapters/ocThreading.html#//apple_ref/doc/uid/TP30001163-CH19-SW1)
+//! - [Source code for
+//! `objc_sync_enter/exit`](https://opensource.apple.com/source/objc4/objc4-551.1/runtime/Accessors.subproj/objc-accessors.mm.auto.html),
+//! otherwise undocumented.
+use crate::{
```
hikari-no-yume: Same stylistic nit I had in the other PR: can you use one line per module, rather than nesting like this? There's nothing wrong with it, it's just not what I do everywhere else.

hikari-no-yume (changes requested):

src/libc/pthread/mutex.rs
abnormalmaps: I think it'd probably be better to keep the mutex state struct separate. I'll move the struct down into `Environment` but keep it as is, unless you'd really prefer it the other way.

hikari-no-yume: To make sure I understand: you want to move this state struct  into `Environment`, rather than having it be in the pthread state struct? And by “keep it as is”, you mean the mutexes hashmap would still be inside the state struct, rather than being directly part of `Environment`? If so, that sounds good to me.

abnormalmaps: I'm not exactly sure it's what you're thinking, but I have the fixes in. If it's not what you're thinking of, I'm open to changing it.

hikari-no-yume: c0a7a58551dc559fb5d45cb36bde886d821251f4 does more or less what I wanted, thanks!

src/environment/mutex.rs
```diff
@@ -0,0 +1,237 @@
...
+        match value {
+            0 => Ok(MutexType::PTHREAD_MUTEX_NORMAL),
+            1 => Ok(MutexType::PTHREAD_MUTEX_ERRORCHECK),
+            2 => Ok(MutexType::PTHREAD_MUTEX_RECURSIVE),
+            _ => Err("Value is not a valid mutex type!"),
+        }
+    }
+}
+pub const PTHREAD_MUTEX_DEFAULT: MutexType = MutexType::PTHREAD_MUTEX_NORMAL;
+/// Custom typedef for readability (the C API just uses `int`)
```
hikari-no-yume: This comment seems to have been stranded, I guess it should be removed now that you've reworked `MutexType`?

abnormalmaps: Thanks for noticing that!

src/libc/pthread/mutex.rs
```diff
@@ -112,23 +93,22 @@ fn pthread_mutex_init(
     let type_ = if !attr.is_null() {
         check_magic!(env, attr, MAGIC_MUTEXATTR);
         let pthread_mutexattr_t { type_, .. } = env.mem.read(attr);
-        assert!(
-            type_ == PTHREAD_MUTEX_NORMAL
-                || type_ == PTHREAD_MUTEX_ERRORCHECK
-                || type_ == PTHREAD_MUTEX_RECURSIVE
-        );
+        let type_ = type_.try_into().unwrap();
```
hikari-no-yume: This does nothing because the `type_` field of the `pthread_mutexattr_t` is a `MutexType`. Since `MutexType` is now an enum rather than an `i32`, you've accidentally introduced undefined behaviour here (the read is effectively a transmute).

I caught this because Clippy warned about the useless `try_into()`. Did you maybe forget to run `cargo clippy` and `cargo fmt` before committing? (I should add these to CI…)

src/environment.rs
```diff
@@ -354,32 +398,69 @@ impl Environment {
...
+        ));
+        log_dbg!(
+            "Thread {} blocking on mutex #{}.",
+            self.current_thread,
+            mutex_id
+        );
+        self.threads[self.current_thread].blocked_by = ThreadBlock::Mutex(mutex_id);
+    }
+
+    pub fn join_with_thread(&mut self, joinee_thread: ThreadID, ptr: MutPtr<MutVoidPtr>) {
```
hikari-no-yume, line 460: This should probably have a comment similar to the one on `block_on_mutex`.

src/dyld.rs
```diff
@@ -137,24 +137,42 @@ fn encode_a32_trap() -> u32 {
...
     return_to_host_routine: Option<GuestFunction>,
+    thread_exit_routine: Option<GuestFunction>,
     constants_to_link_later: Vec<(MutPtr<ConstVoidPtr>, &'static HostConstant)>,
 }
 
 impl Dyld {
     /// We reserve this SVC ID for invoking the lazy linker.
-    const SVC_LAZY_LINK: u32 = 0;
+    pub const SVC_LAZY_LINK: u32 = 0;
+    /// We reserve this SVC ID as the return for function entry.
```
hikari-no-yume: “function entry” sounds a bit too close to what return-to-host does, maybe “the thread start routine”?

hikari-no-yume (changes requested):
I'm very close to merging this.

src/libc/pthread/mutex.rs
abnormalmaps: Thanks for catching this :D

> I caught this because Clippy warned about the useless try_into(). Did you maybe forget to run cargo clippy and cargo fmt before committing? (I should add these to CI…)

Apparently I didn't. 

It might nice to have a derive that checks members for `SafeRead/Write`ability to make this kind of mistake harder. I might get around to it (or might not).

abnormalmaps:
Sorry for being slow to get back to this! I hope the fixes are acceptable to you. I'll probably take a pass over this myself later.

src/environment.rs
```diff
@@ -450,6 +451,12 @@ impl Environment {
         self.threads[self.current_thread].blocked_by = ThreadBlock::Mutex(mutex_id);
     }
 
+    /// Blocks the current thread until the thread given finishes, writing it's return value to ptr
+    /// (if non-null).
+    /// Note that there are no protections against joining with a deattached thread, joining a thread with itself,
```
hikari-no-yume, line 456: Nit: deattached -> detached.

hikari-no-yume: Oh also, FWIW I prefer it if comments don't go past the 80-column mark. I should really set up something to detect and enforce this kind of thing, it's a waste of time to enforce this manually.

hikari-no-yume:
> It might nice to have a derive that checks members [of structs] for `SafeRead`[…] to make this kind of [UB] mistake harder. I might get around to it (or might not).

Ah, I guess that would be useful. So far touchHLE has no procedural macros at all, I've stubbornly used macros-by-example for everything. One downside is there's no derive macros, and the macros I do have for implementing traits have no way to statically check these kinds of invariants, which is a bit frustrating. Though, how would a procedural macro check this? It can't see the actual definitions of the types of the members, just the definition of the struct itself, I think. I guess it could generate code that reads each member individually, and that would fail to compile, but it'd be less efficient than the current `unsafe impl SafeRead for EntireStruct {}` trick.

hikari-no-yume:
Ooh actually, a derive macro could generate code like this:

```rust
struct A;
struct B;
struct C;
unsafe impl SafeRead for A {}
unsafe impl SafeRead for B {}

struct ABC {
    a: A,
    b: B,
    c: C,
}

unsafe impl SafeRead for ABC where A: SafeRead, B: SafeRead, C: SafeRead {}
```

[rustc seems to do what I want](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f1b1569b0c03db441cf0ac8809453d04):

```
error[[E0277]](https://doc.rust-lang.org/stable/error_codes/E0277.html): the trait bound `C: SafeRead` is not satisfied
  --> src/lib.rs:15:62
   |
15 | unsafe impl SafeRead for ABC where A: SafeRead, B: SafeRead, C: SafeRead {}
   |                                                              ^^^^^^^^^^^ the trait `SafeRead` is not implemented for `C`
   |
```

Though I think my preference would be to use a macro-by-example to generate this, rather than a procedural macro… hmm.

src/environment.rs
```diff
@@ -8,26 +8,38 @@
 //! Unlike its siblings, this module should be considered private and only used
 //! via the re-exports one level up.
 
+pub mod mutex;
+
+use crate::abi::GuestRet;
+
+use crate::environment::mutex::{host_mutex_is_locked, host_mutex_relock_unblocked};
```
hikari-no-yume, line 15: Nit: might want to consolidate this with the `use self::mutex::` below and just write `use mutex::`. Also remove the double new lines here.

src/objc/synchronization.rs
```diff
@@ -0,0 +1,103 @@
...
+    host_mutex_destroy, host_mutex_init, host_mutex_lock, host_mutex_unlock, MutexType,
+};
+use crate::Environment;
+
+use super::id;
+
+/// Backing function of @synchronized block entry.
+/// This function is entirely undocumented, with
+/// [source code provided](https://opensource.apple.com/source/objc4/objc4-551.1/runtime/objc-sync.h.auto.html).
+pub(super) fn objc_sync_enter(env: &mut Environment, obj: id) -> u32 {
```
hikari-no-yume, line 27: Nit: C `int` corresponds to `i32`. (I should add typedefs for the C types already…)

src/objc/synchronization.rs
```diff
@@ -0,0 +1,103 @@
...
+        host_mutex_lock(env, mutex_id).unwrap();
+        env.objc.sync_mutexes.insert(obj, mutex_id);
+    }
+    0u32 // OK
+}
+
+/// Backing function of @synchronized block exit.
+/// This function is entirely undocumented, with
+/// [source code provided](https://opensource.apple.com/source/objc4/objc4-551.1/runtime/objc-sync.h.auto.html).
+pub(super) fn objc_sync_exit(env: &mut Environment, obj: id) -> u32 {
```
hikari-no-yume, line 51: Ditto here.

src/environment/mutex.rs
```diff
@@ -0,0 +1,236 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+//! Internal mutex interface.
+
+use std::collections::HashMap;
+use std::num::NonZeroU32;
+
+use crate::libc::errno::{EBUSY, EDEADLK, EPERM};
+use crate::libc::pthread::mutex::HostMutexId;
```
hikari-no-yume, line 12: Hmm, wouldn't it make more sense for the definition of `HostMutexId` to be here rather than in pthreads, now? It could also just be called `MutexId` in that case; the host/guest distinction only matters within pthreads, which deals with both.

hikari-no-yume:
Since there seems to be a slow review cycle time here (no pressure, I'm sure you have limited time and better things to do!), I've decided to just merge it as is and make some tweaks myself on trunk.

PR #134, opened 2023-07-31
Support for Donut games

ciciplusplus:
Working games:

- Cat Physics ([video](https://twitter.com/ciciplusplus/status/1683519097386967040))
Screenshot 2023-07-31 at 20 51 51

- Spikey's Bounce Around
Screenshot 2023-07-31 at 20 41 43

- Traffic Rush
Screenshot 2023-07-31 at 20 40 17

- Castle Smasher
Screenshot 2023-07-31 at 20 38 31

- Rat on a Scooter XL
Screenshot 2023-07-24 at 23 39 13

- Rat on a Scateboard
Screenshot 2023-07-31 at 20 35 33

- Ruberth's Kick n' Fly
Screenshot 2023-07-31 at 20 49 38

- Titanic Rescue
Screenshot 2023-07-31 at 00 06 06

- Icy Escort
Screenshot 2023-07-31 at 00 08 34

P.S. More iOS 2.x games from the same publisher may work as well, I just don't happen to have them to test


src/fs.rs
```diff
@@ -507,6 +507,15 @@ impl Fs {
         &self.home_directory
     }
 
+    pub fn change_current_directory(&mut self, path: &str) {
+        if path.starts_with('/') {
+            self.current_directory = GuestPathBuf::from(GuestPath::new(&path));
+        } else {
+            self.current_directory = self.current_directory.join(path);
```
hikari-no-yume: The current `GuestPath::join` implementation doesn't handle empty, `.` and `..` path components neatly. Using `resolve_path` would not only avoid that problem, but it also would let you remove the `path.starts_with('/')` check.

src/fs.rs
```diff
@@ -507,6 +507,15 @@ impl Fs {
...
+        }
+        assert!(self.exists(&self.current_directory));
```
hikari-no-yume: Maybe make this into the return value? `Result<(), ()>` is a bit of a weird type but would clearly signal that it can fail.

src/libc/stdio.rs
```diff
@@ -152,6 +152,17 @@ fn fclose(env: &mut Environment, file_ptr: MutPtr<FILE>) -> i32 {
     }
 }
 
+fn fsetpos(env: &mut Environment, file_ptr: MutPtr<FILE>, pos: ConstPtr<i64>) -> i32 {
+    let res = i32::try_from(env.mem.read(pos)).unwrap();
+    fseek(env, file_ptr, res, SEEK_SET)
```
hikari-no-yume: It might be better to use `posix_io::lseek` so there isn't a 64-bit/32-bit conversion.

Also, you should add a type alias for `i64` here. According to the manual, the type is called `fpos_t`.

(These comments apply to `fgetpos` also.)

src/fs.rs
ciciplusplus: Done

src/fs.rs
ciciplusplus: Done

src/libc/stdio.rs
ciciplusplus: Done for both

src/fs.rs
```diff
@@ -558,16 +558,17 @@ impl Fs {
...
         let Some(FsNode::Directory { children, .. }) = self.lookup_node(path.as_ref()) else {
             return Err(());
         };
-        Ok(children.keys().map(|name| name.as_str()))
+        let mut paths = Vec::new();
+        let iter = children.keys().map(|name| name.as_str());
+        for name in iter {
+            paths.push(GuestPathBuf::from(GuestPath::new(name)));
+        }
+        Ok(paths)
```
hikari-no-yume: Hmm, why did you change this? There's nothing wrong with it in principle, but having an iterator is usually better, and you could use `.map` to convert to strings in the calling function instead.

src/frameworks/core_foundation/time.rs
```diff
@@ -5,17 +5,94 @@
...
+            seconds: GuestArg::from_regs(&regs[5..7]),
+        }
+    }
+    fn to_regs(self, regs: &mut [u32]) {
+        self.year.to_regs(&mut regs[0..1]);
+        self.month.to_regs(&mut regs[1..2]);
+        self.day.to_regs(&mut regs[2..3]);
+        self.hours.to_regs(&mut regs[3..4]);
+        self.minutes.to_regs(&mut regs[4..5]);
+        self.seconds.to_regs(&mut regs[5..7]);
```
hikari-no-yume: Wait, really? Did you verify this with Ghidra?

src/audio/openal_soft_wrapper/lib.rs
hikari-no-yume, line 1: Can you give a better title to this commit? :p

hikari-no-yume:
I've cherry-picked a few of these now:

| original                                 | on trunk                                 | notes                         |
| ---------------------------------------- | ---------------------------------------- | ----------------------------- |
| e97d65122d9630ebea8a0cbf1182feb8ae6a1629 | 298d5aad6fe6332f1c6661ca559bd780ac64edd7 | merge conflict (removed test) |
| 251003fd139d381930d55dff57ddf80eadb81d19 | e768e1e893cc8a4774a1c124f802ffd53f25e8f6 |                               |
| bd573d7d06000d798ee28d945c9486e8216c9a46 | d89b39247ed10f9e946650599af3d10eddb9f2d6 | merge conflict                |
| b6c5498a8f533b25420e66d71e1d6757b6c05877 | 0d8f2a68787f9cca223e2d8e9a0dee61b4723a0b | merge conflict                |
| 67a598cb87b47725df6c1bcd679c66d6c63dcad8 | a2430fc42a6653124a8adb95dd64820eb537c3a5 | merge conflict (removed test) |
| 452e4670c2e7f9f79494bdca79dd20a2543072d0 | 94575367f3d2ab611702e34fed14bf848cb363b5 |                               |

src/fs.rs
hikari-no-yume: Oh, was it to make it consistent with `enumerate_recursive`? That is sort of reasonable in its own way, but that function returns a `Vec<GuestPathBuf>` because it has to provide fully-qualified paths, while this function doesn't.

src/frameworks/core_foundation/time.rs
ciciplusplus: Not really... How can I do that?

src/audio/openal_soft_wrapper/lib.rs
ciciplusplus: Done ;)

src/fs.rs
ciciplusplus: I don't remember the exact reasoning as it was more than month ago :)
I saw that it was marked as [allow dead_code], so I took a liberty to modify the interface

ciciplusplus: I changed signature back to the previous one, please check

src/frameworks/core_foundation/time.rs
hikari-no-yume: Uh, well, you look at a function call involving the type and try to figure out the pattern of registers involved.

This grabbed my attention because the in-memory struct is seemingly quite compact (16 bytes), but if it's passed like this in registers or on the stack, then it's not so efficient (28 bytes). I'm not sure, but it might make sense if it has the same representation in both cases.

sheva22222:
try testing these games: Spikey_1.12 Traffic_Rush_1.32_Greatest_Ever Skateboard_1.10 Rat_On_A_Snowboard_1.01 Cat_Physics_1.01 Jailhouse_Jack_1.03 Jungle_Crash_Land_1.34 Cave_Bowling_1.14

src/frameworks/core_foundation/time.rs
ciciplusplus: I don't think this is actually matter - as soon as from_regs and to_regs are consistent, it would work.
Anyway, I changed this code to use only 4 registers.

I also confirmed that it works with small test app (see screenshot). Both cases (with 7 and with 4 registers) are working correctly.
Screenshot 2023-08-29 at 20 26 28

src/fs.rs
hikari-no-yume: Looks good to me, thanks for changing it.

src/frameworks/core_foundation/time.rs
hikari-no-yume: > [if] from_regs and to_regs are consistent, it would work.

When passing data between touchHLE's own implementations, sure, but not when the struct is directly accessed by the app…

> I also confirmed that it works with small test app (see screenshot). Both cases (with 7 and with 4 registers) are working correctly.

Uhh… I'm scared now, I don't understand how you'd be able to write that C code and have it work for both 7 and 4 registers with the same results…

ciciplusplus: Ok, i think i understand, my last test program has not involved host function calls involving CFGregorianDate, that's why it's execution is independent of CFGregorianDate's GuestArg definition... Sorry for the noise.

ciciplusplus: @hikari-no-yume  Ok, so... Finally I don't need to define GuestArg at all, only GuestRet :) So, I just removed GuestArg for CFGregorianDate from this PR.

Anyway, this is last version of test that I'm using:

```
int main(int argc, const char * argv[]) {
    time_t unixTime = time(NULL);
    CFAbsoluteTime at = CFAbsoluteTimeGetCurrent();
    printf("unix %d, abs %f, diff %d\n", unixTime, at, (int)ceil(unixTime - at));
    CFGregorianDate date = CFAbsoluteTimeGetGregorianDate(at, nil);
    printf("year %d, month %d, day %d, hour %d, minute %d, second %f\n",
        date.year, date.month, date.day, date.hour, date.minute, date.second);
    return 0;
}
```

Two things to notice:
1. Current CFAbsoluteTimeGetCurrent() returns time from startup of the programs, not a correct reference date. I haven't changed that.
2. seconds of CFGregorianDate do not have fractional part as i'm re-using timestamp_to_calendar_date() function.

Let me know if those need to be addressed before the merge.

Screenshot 2023-08-30 at 12 37 26

ciciplusplus:
> try testing these games: Spikey_1.12 Traffic_Rush_1.32_Greatest_Ever Skateboard_1.10 Rat_On_A_Snowboard_1.01 Cat_Physics_1.01 Jailhouse_Jack_1.03 Jungle_Crash_Land_1.34 Cave_Bowling_1.14

I don't have those versions at hand to test, but if those are iOS 2 versions, there is high chance they would work as well :)

src/frameworks/core_foundation/time.rs
hikari-no-yume: > Current CFAbsoluteTimeGetCurrent() returns time from startup of the programs, not a correct reference date. I haven't changed that.

Oh right, it's meant to be wall-clock time based on some weird Apple epoch, right? I think the current implementation was just some oversight on my part.

Uhh, since you're adding date conversion functions, I'd rather you use this opportunity to fix it, since otherwise all the dates are just nonsense. You can use Rust's `SystemTime::now()`, `SystemTime::UNIX_EPOCH` and the `Duration` functions that convert to/from floats. I suggest defining a constant like `const APPLE_EPOCH: SystemTime = UNIX_EPOCH + …;` with a precalculated offset.

Also, hmm… I guess `timestamp_to_calendar_date` should be changed to be 64-bit at some point, but I won't make you do that. You can add a TODO comment if you want.

> seconds of CFGregorianDate do not have fractional part as i'm re-using timestamp_to_calendar_date() function.

There's no complexity to extracting seconds, you can just call the `.fract()` method on the f64.

hikari-no-yume: Oh also, have you considered adding this test to TestApp? While we have to provide our own type definitions, if the date struct is declared with equivalent types to those in Apple's headers, it still serves as a useful test.

ciciplusplus: > There's no complexity to extracting seconds, you can just call the .fract() method on the f64.

timestamp_to_calendar_date() returns tm struct with seconds defined as i32

> Oh also, have you considered adding this test to TestApp?

I have considered that, but given the current state of this PR this is cumbersome - I would need to recreate the PR from trunk with only commits which are not merged. I can do that, but this would screw up this PR history.

ciciplusplus: > Uhh, since you're adding date conversion functions, I'd rather you use this opportunity to fix it, since otherwise all the dates are just nonsense. You can use Rust's SystemTime::now(), SystemTime::UNIX_EPOCH and the Duration functions that convert to/from floats. I suggest defining a constant like const APPLE_EPOCH: SystemTime = UNIX_EPOCH + …; with a precalculated offset.

Done.
(rust complained when i defined it as const because of non-const fn usage, so i defined it as a function finally.)

hikari-no-yume: > Done.  
> (rust complained when i defined it as const because of non-const fn usage, so i defined it as a function finally.)

Seems good to me, thanks. Nitpick: touchHLE only uses CamelCase functional names where Apple's APIs do. In this case, it's an internal function, so `snake_case` is preferred.

> > There's no complexity to extracting seconds, you can just call the .fract() method on the f64.
>
> timestamp_to_calendar_date() returns tm struct with seconds defined as i32

Oh, what I mean is you can bypass timestamp_to_calendar_date and extract the seconds directly from the float. I made a mistake though: `.frac()` would be the fraction of a second, not the number of seconds in a minute, but you could add it to the result from that function.

Neither is a big issue, I can fix it myself on trunk if necessary.

ciciplusplus: > > Done.
> > (rust complained when i defined it as const because of non-const fn usage, so i defined it as a function finally.)
> 
> Seems good to me, thanks. Nitpick: touchHLE only uses CamelCase functional names where Apple's APIs do. In this case, it's an internal function, so `snake_case` is preferred.

done :)

hikari-no-yume:
I cherry-picked some more commits:

| original                                 | on trunk                                 | notes                         |
| ---------------------------------------- | ---------------------------------------- | ----------------------------- |
| b8a0ddf2b6cbc4655fe573b2f46fec8079ee97c0 | ce09edd1e788267862553a3d99e6b15dcf85d5ff |                               |
| 63b9cc29b8c78c20d7c71ca792065273beb93626 | a6691f9f9f201f53448d21fc168d92fb683162d8 |                               |
| 795b8c9ec749045d10d32291651ab4c38e88bc25 | 9f4208af4c6bec1e4e1f11fd2546914a4732f83e | merge conflict (removed test) |
| 0f326e2345fdbaaa06ad9a9eb510a7ecc3728175 | bbfe22e96b34dfee2f00110ea728efa24b9594a7 |                               |
| 9886661320b13b69dbf04279a0ae0a3b1b0eb484 | fb128e870e8f8eae16bba6889e5e689d3ce483f4 |                               |

I made some small changes to the NSFileManager stuff on trunk here: d4743fbdc8973c5efd9c1affb6475de89908e05c.

Instead of cherry-picking your chdir(), I did my own here: c8e1eb1857ff6d60f698d63328c0e8185272d9df.

Also I didn't mention it before, but there were small fixes for some earlier cherry-picks also: 8dbd5c46a7d7a11f5c65b8061de060017a7acd4b e3fd33ce34f7f2107b82a56b300c65e82994559f.

I think the only things left now are 6009c34d0c1d45c73b34c8a8302f68fafecfacd7, 9f57d4d7cf64eafc3e7296e67f43291a2041c235, 0f6e3345994708f67ae0f44ea6af4de3c3489320, 21c4903aa60de6874a63f569a6d03a1ebccf22e7 and 162a7487050ef9b42aa8f8289fe3b393c47992b7.

src/libc/stdio.rs
```diff
@@ -152,6 +155,28 @@ fn fclose(env: &mut Environment, file_ptr: MutPtr<FILE>) -> i32 {
     }
 }
 
+fn fsetpos(env: &mut Environment, file_ptr: MutPtr<FILE>, pos: ConstPtr<fpos_t>) -> i32 {
+    let FILE { fd } = env.mem.read(file_ptr);
+
+    let res = posix_io::lseek(env, fd, env.mem.read(pos), SEEK_SET);
+    assert_ne!(res, -1);
```
hikari-no-yume: Any reason you have an assert in these two functions? There is already logging in `lseek`. I think you can just return `if res == -1 { -1 } else { 0 }` in both of these.

src/libc/stdio.rs
```diff
@@ -27,6 +27,9 @@ struct FILE {
 }
 unsafe impl SafeRead for FILE {}
 
+#[allow(non_camel_case_types)]
+type fpos_t = i64;
```
hikari-no-yume: Apparently this is just an alias of `off_t` so maybe it should be declared that way.

src/libc/stdio.rs
ciciplusplus: Done

src/libc/stdio.rs
ciciplusplus: Done

hikari-no-yume:
I have now cherry-picked bf1768c284140da2d5213f0606af4d101b5a5564 as 841d34a1e1c45ad72a81a1bbe2ca797735056b0c. 

I did my own `initWithString:` implementation (846c95dfdd702ca7a8e4a18148be463b581edaec) and `calloc` fix (b47b787f74fd46cdb3f6035cde1545af4d19d1ca).

I'll probably do my own `alcSuspendContext` one since it'll merge conflict with my own change that adds `alGetListenerf`, and I think it'd be good to have symmetry and implement the opposite function (edit: done: c04ed7a33dbdfa3f48a49c55b5383ec44d96ff1a).

The only one that remains now is strcpy chk, I need to research that a little.

hikari-no-yume:
Cherry-picked a0142207c201257f5663924ec6d1aff178d24528 as 449b5f63367379b2f47743b5768bee7ea1fa3924. I'm not very happy with it but I can't be bothered to do it properly for now, so I made an issue: https://github.com/hikari-no-yume/touchHLE/issues/145.

With that, I think I can close this!

ciciplusplus:
🥳 thanks!

sheva22222:
games that work on my pc: cat physics, spikes, traffic jams, fastlane street racing (favorite version), crash bandicoot, scooter xl, castle(touchhle 0.2.0)

PR #154, opened 2023-09-06
Add runtime type checking for host-to-host messages in debug builds

### LINKED ISSUE #140: Type checking for host message sends
Currently message sends in host code (`msg![env; …]`) have no type safety. I'm quite vigilant about using the correct types, but I can't expect everyone to immediately see what kinds of usages are risky, and in any case mistakes will always be made. It'd be good to do basic checks: are the float and integer parameters the right size, for example? They'd have to be dynamic, so maybe this should only be done in debug mode.

abnormalmaps:
Partially fixes #140. 

It should be doable to check host to guest calls as well, but that would involve parsing the type string and I didn't really feel like doing that now :P. Some of the weird looking commits are to make the `msg!` calls work again, and I also took the liberty of macro-ifying the HostIMP calls.

### DIFF
--- a/src/frameworks/foundation.rs
+++ b/src/frameworks/foundation.rs
@@ -55,7 +55,7 @@ pub struct State {
 pub type NSInteger = i32;
 pub type NSUInteger = u32;
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 #[repr(C, packed)]
 pub struct NSRange {
     pub location: NSUInteger,
--- a/src/frameworks/foundation/ns_string.rs
+++ b/src/frameworks/foundation/ns_string.rs
@@ -18,10 +18,10 @@ use crate::frameworks::uikit::ui_font::{
 };
 use crate::fs::GuestPath;
 use crate::mach_o::MachO;
-use crate::mem::{guest_size_of, ConstPtr, Mem, MutPtr, Ptr, SafeRead};
+use crate::mem::{guest_size_of, ConstPtr, Mem, MutPtr, MutVoidPtr, Ptr, SafeRead};
 use crate::objc::{
-    autorelease, id, msg, msg_class, nil, objc_classes, retain, Class, ClassExports, HostObject,
-    NSZonePtr, ObjC,
+    autorelease, id, msg, msg_class, msg_class_unchecked, nil, objc_classes, retain, Class,
+    ClassExports, HostObject, NSZonePtr, ObjC,
 };
 use crate::Environment;
 use std::borrow::Cow;
@@ -505,12 +505,12 @@ pub const CLASSES: ClassExports = objc_classes! {
 - (ConstPtr<u8>)UTF8String {
     // TODO: avoid copying
     let string = to_rust_string(env, this);
-    let c_string = env.mem.alloc_and_write_cstr(string.as_bytes()).cast_const();
+    let c_string = env.mem.alloc_and_write_cstr(string.as_bytes());
     let length: NSUInteger = (string.len() + 1).try_into().unwrap();
     // NSData will handle releasing the string (it is autoreleased)
-    let _: id = msg_class![env; NSData dataWithBytesNoCopy:c_string
+    let _: id = msg_class![env; NSData dataWithBytesNoCopy:(c_string.cast_void())
                                                     length:length];
-    c_string
+    c_string.cast_const()
 }
 
 - (id)stringByTrimmingCharactersInSet:(id)set { // NSCharacterSet*
@@ -933,7 +933,12 @@ pub fn get_static_str(env: &mut Environment, from: &'static str) -> id {
     if let Some(&existing) = State::get(env).static_str_pool.get(from) {
         existing
     } else {
-        let new = msg_class![env; _touchHLE_NSString_Static alloc];
+        // Since this gets called during late linking (before the cpu is properly set
+        // up), we can't have it use the stack for debug info. [NSObject alloc:] calls
+        // [this allocWithZone:] internally, so it would also have to be unchecked. Calling
+        // allocWithZone: directly avoids this.
+        let new =
+            msg_class_unchecked![env; _touchHLE_NSString_Static allocWithZone:(MutVoidPtr::null())];
         *env.objc.borrow_mut(new) = StringHostObject::Utf8(Cow::Borrowed(from));
         State::get(env).static_str_pool.insert(from, new);
         new
--- a/src/frameworks/foundation/ns_timer.rs
+++ b/src/frameworks/foundation/ns_timer.rs
@@ -79,7 +79,7 @@ pub const CLASSES: ClassExports = objc_classes! {
                             selector:(SEL)selector
                             userInfo:(id)user_info
                              repeats:(bool)repeats {
-    let timer = msg![env; this timerWithTimeInterval:interval
+    let timer: id = msg![env; this timerWithTimeInterval:interval
                                               target:target
                                             selector:selector
                                             userInfo:user_info
--- a/src/mem.rs
+++ b/src/mem.rs
@@ -91,6 +91,10 @@ impl<T, const MUT: bool> Ptr<T, MUT> {
     pub fn is_null(self) -> bool {
         self.to_bits() == 0
     }
+
+    pub fn cast_void(self) -> Ptr<std::ffi::c_void, MUT> {
+        self.cast::<std::ffi::c_void>()
+    }
 }
 
 impl<T> ConstPtr<T> {
--- a/src/objc/messages.rs
+++ b/src/objc/messages.rs
@@ -11,6 +11,10 @@
 //! - Mike Ash's [objc_msgSend's New Prototype](https://www.mikeash.com/pyblog/objc_msgsends-new-prototype.html)
 //! - Peter Steinberger's [Calling Super at Runtime in Swift](https://steipete.com/posts/calling-super-at-runtime/) explains `objc_msgSendSuper2`
 
+use std::any::{Any, TypeId};
+use std::collections::hash_map::DefaultHasher;
+use std::hash::{Hash, Hasher};
+
 use super::{id, nil, Class, ObjC, IMP, SEL};
 use crate::abi::{CallFromHost, GuestRet};
 use crate::mem::{ConstPtr, MutVoidPtr, SafeRead};
@@ -23,13 +27,22 @@ use crate::Environment;
 /// variant may have additional arguments to be forwarded (or rather, left
 /// untouched) by `objc_msgSend` when it tail-calls the method implementation it
 /// looks up. This is invisible to the Rust type system; we're relying on
-/// [crate::abi::CallFromGuest] here.
+/// [crate::abi::CallFromGuest] here. To provide (limited) typechecking support,
+/// messages that are called by host functions and received by host classes can have
+/// their parameters checked at runtime, which is done transparently through the [msg]
+/// macro (and friends).
 ///
 /// Similarly, the return value of `objc_msgSend` is whatever value is returned
 /// by the method implementation. We are relying on CallFromGuest not
 /// overwriting it.
 #[allow(non_snake_case)]
-fn objc_msgSend_inner(env: &mut Environment, receiver: id, selector: SEL, super2: Option<Class>) {
+fn objc_msgSend_inner(
+    env: &mut Environment,
+    receiver: id,
+    selector: SEL,
+    super2: Option<Class>,
+    type_id: Option<u64>,
+) {
     if receiver == nil {
         // https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjectiveC/Chapters/ocObjectsClasses.html#//apple_ref/doc/uid/TP30001163-CH11-SW7
         log_dbg!("[nil {}]", selector.as_str(&env.mem));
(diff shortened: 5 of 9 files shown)


hikari-no-yume:
Whoa, this is really cool! I wasn't expecting anyone other than me to attempt this, honestly. Thank you!

Apple's Objective-C does actually have type info strings for methods, which touchHLE currently ignores, so it would be nice if we could use that instead of Rust's `type_id`, but this is still a big improvement as it is. I don't want to make you rewrite it, I think someone (possibly me) could do it in follow-up work.

src/frameworks/foundation/ns_string.rs
```diff
@@ -933,7 +933,10 @@ pub fn get_static_str(env: &mut Environment, from: &'static str) -> id {
     if let Some(&existing) = State::get(env).static_str_pool.get(from) {
         existing
     } else {
-        let new = msg_class![env; _touchHLE_NSString_Static alloc];
+        // Using allocWithZone: instead of alloc: allows us to avoid having to leave alloc: as
+        // unchecked.
```
hikari-no-yume: Could you explain what the type error here, or in `alloc`, is? (By the way, there's no colon at the end of its name, it has no argument.) It's not obvious to me what would be wrong.

src/mem.rs
```diff
@@ -91,6 +91,10 @@ impl<T, const MUT: bool> Ptr<T, MUT> {
     pub fn is_null(self) -> bool {
         self.to_bits() == 0
     }
+
+    pub fn cast_void(self) -> Ptr<std::ffi::c_void, MUT> {
```
hikari-no-yume, line 95: This is a good idea and fits the convention of never importing `std::ffi::c_void`, thanks for adding it.

src/objc/messages.rs
```diff
@@ -249,6 +399,53 @@ where
...
 macro_rules! msg {
+    [$env:expr; $receiver:tt $name:ident $(: $arg1:tt)?
+                             $($namen:ident: $argn:tt)* $(,$va_argn:ident)*] => {
+        {
+            let sel = $crate::objc::selector!($($arg1;)? $name $(, $namen)*);
+            let sel = $env.objc.lookup_selector(sel)
+                .expect("Unknown selector");
+            if cfg!(debug_assertions)
+                && $env.cpu.regs_mut()[$crate::cpu::Cpu::SP]
+                    > $crate::mem::guest_size_of::<$crate::objc::objc_super_debug>()
```
hikari-no-yume: If the stack pointer has gotten low enough that this check would make sense, it's already overwritten the entire app binary and more. I guess this functions as an overflow check, but since you're only doing this in debug builds, the default Rust overflow checks will apply anyway. So I would suggest just removing this.

src/objc/messages.rs
```diff
@@ -373,3 +649,16 @@ pub fn autorelease(env: &mut Environment, object: id) -> id {
     }
     msg![env; object autorelease]
 }
+
+/// Utility function to generate type ids used in debug type checking.
+pub fn generate_typeid<T>(_arg: T) -> u64
```
hikari-no-yume: For consistency:

```diff
-pub fn generate_typeid<T>(_arg: T) -> u64
+pub fn generate_type_id<T>(_arg: T) -> u64
```

src/frameworks/foundation/ns_string.rs
abnormalmaps: I should probably have specified, it's because this gets called during late linking, and the stack pointer is set to zero at that point. If you call `alloc:`, it dips into NSObject, so i'd have to set that as unchecked as well.

src/objc/messages.rs
abnormalmaps: Fair enough. I had this for the late linking part (up above), and thought the error message was a bit unintuitive. I can change it back though.

src/mem.rs
abnormalmaps: No problem!

src/objc/messages.rs
abnormalmaps: With that said, I don't think it makes sense for `msg_super!` since that won't work during linking either way, so I guess that can go.

src/frameworks/foundation/ns_string.rs
hikari-no-yume: Ahh. The stack setup should probably be done earlier then.

src/objc/messages.rs
abnormalmaps: Got rid of it entirely, I don't think it'll be a common enough issue to matter.

hikari-no-yume:
I think I'd like to get rid of all the `msg_unchecked!` and `msg_class_unchecked!` uses and do away with that macro. I can't think of a case where it would actually be justified to use it. To that end:

- `prep_stack_for_start` can be moved before the call to `Dyld::do_late_linking` in `Environment::new`. (Note that the comment in `Environment::new_without_app` should also be moved.) Then there's no need for the special case when constructing `NSString`.
- The `NSDictionary` convenience method can be refactored along the lines of 204d05e7c7e4060cfd66f3be4293e4de4e647fcb, removing the hack.

Does that sound good to you?

src/frameworks/foundation.rs
```diff
@@ -55,7 +55,7 @@ pub struct State {
 pub type NSInteger = i32;
 pub type NSUInteger = u32;
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
hikari-no-yume, line 58: Was this an accidental change or something?

src/frameworks/foundation/ns_timer.rs
```diff
@@ -79,7 +79,7 @@ pub const CLASSES: ClassExports = objc_classes! {
                             selector:(SEL)selector
                             userInfo:(id)user_info
                              repeats:(bool)repeats {
-    let timer = msg![env; this timerWithTimeInterval:interval
+    let timer: id = msg![env; this timerWithTimeInterval:interval
```
hikari-no-yume, line 82: Please fix the indentation of the subsequent lines if you're going to change this. It should already be inferred as `id` due to the return type so just reverting it is also an option.

src/objc/messages.rs
```diff
@@ -134,7 +152,42 @@ fn objc_msgSend_inner(env: &mut Environment, receiver: id, selector: SEL, super2
 /// Standard variant of `objc_msgSend`. See [objc_msgSend_inner].
 #[allow(non_snake_case)]
 pub(super) fn objc_msgSend(env: &mut Environment, receiver: id, selector: SEL) {
-    objc_msgSend_inner(env, receiver, selector, /* super2: */ None)
+    objc_msgSend_inner(env, receiver, selector, /* super2: */ None, None)
+}
+
+#[repr(C, packed)]
+/// A pointer to this struct replaces the normal receiver parameter for
+/// `objc_msgSend(stret)_debug` and [msg_send] when debug assertions are on.
```
hikari-no-yume, line 160: That should be `objc_msgSend(_stret)_debug` right?

Also, the comment doesn't seem to match the implementation, isn't it only used for `msg_send_debug`?

hikari-no-yume: Oh, also, you could just write `[objc_msgSend_debug], [objc_msgSend_stret_debug] and [msg_send]`.

src/objc/messages.rs
```diff
@@ -186,17 +261,57 @@ pub(super) fn objc_msgSendSuper2(
...
+        selector,
+        /* super2: */ Some(class),
+        None,
+    )
+}
+
+#[repr(C, packed)]
+/// Variant of [objc_super] that enables typechecking from host calls.
+/// A pointer to this struct replaces the normal receiver parameter for
+/// `objc_msgSendSuper2_debug` and [msg_send_super2_debug].
```
hikari-no-yume, line 276: You may want to make `objc_msgSendSuper2_debug` into a link.

src/objc/messages.rs
```diff
@@ -249,6 +399,46 @@ where
...
 #[macro_export]
 macro_rules! msg {
+    [$env:expr; $receiver:tt $name:ident $(: $arg1:tt)?
+                             $($namen:ident: $argn:tt)* $(,$va_argn:ident)*] => {
+        {
+            let sel = $crate::objc::selector!($($arg1;)? $name $(, $namen)*);
+            let sel = $env.objc.lookup_selector(sel)
+                .expect("Unknown selector");
+            if cfg!(debug_assertions)
+            {
```
hikari-no-yume, line 409: Nitpick: this should be on the previous line so it doesn't look like it's a separate block.

src/objc/messages.rs
```diff
@@ -298,7 +488,62 @@ macro_rules! msg_super {
             let sel = $crate::objc::selector!($($arg1;)? $name $(, $namen)*);
             let sel = $env.objc.lookup_selector(sel)
                 .expect("Unknown selector");
+            if cfg!(debug_assertions)
+            {
```
hikari-no-yume, line 492: Ditto.

src/objc/messages.rs
```diff
@@ -373,3 +637,16 @@ pub fn autorelease(env: &mut Environment, object: id) -> id {
...
     msg![env; object autorelease]
 }
+
+/// Utility function to generate type ids used in debug type checking.
+pub fn generate_type_id<T>(_arg: T) -> u64
+where
+    T: Any,
+{
+    // Since we can't pass the TypeId across ABI boundaries, we need to hash it.
+    // Instead, we can hash the TypeId and pass that across.
```
hikari-no-yume, line 647: The second line seems redundant.

src/objc/methods.rs
```diff
@@ -27,91 +29,49 @@ pub enum IMP {
...
-    P5: GuestArg,
-{
+
+macro_rules! impl_HostIMP {
+    ($($P:ident),* ) => {
+
+        impl<R, $($P),*> HostIMP for fn(&mut Environment, id, SEL, $($P),*) -> R where
+            R: GuestRet,
+            $($P: GuestArg + 'static,)*
+            {
```
hikari-no-yume, line 46: This should have one less level of indentation?

hikari-no-yume:
I hope the little things aren't overwhelming. I'm pretty happy with this overall!

src/frameworks/foundation/ns_timer.rs
abnormalmaps: For some reason, it doesn't infer the type properly (on rustc 1.71.1). I'll fix the indentation though.

src/objc/messages.rs
```diff
@@ -249,6 +399,46 @@ where
...
 macro_rules! msg {
+    [$env:expr; $receiver:tt $name:ident $(: $arg1:tt)?
+                             $($namen:ident: $argn:tt)* $(,$va_argn:ident)*] => {
+        {
+            let sel = $crate::objc::selector!($($arg1;)? $name $(, $namen)*);
+            let sel = $env.objc.lookup_selector(sel)
+                .expect("Unknown selector");
+            if cfg!(debug_assertions)
+            {
+                let type_id = $crate::objc::generate_type_id(($($arg1.clone(),)? $($argn.clone(),)*));
```
hikari-no-yume, line 410: Now that I look a bit closer at this, I wonder if it could cause problems:

- The types being used to generate the type ID are not the actual types of the arguments, but the results of cloning them. So something weird could happen if there's references involved, maybe?
- There's no checking for return values.
- Cloning these values only to throw them away seems inefficient.
- While the current types implementing `GuestArg` are all trivially copyable (or could be made to be), maybe they won't be in future.

I think it would be cleaner if the type ID could be generated directly from the types, rather than cloned values. Obviously, we don't have access to them here in the macro invocation, but I think it could be done within `msg_send_debug`? If you add a type bound to it that uses `HostIMP`, and move the `objc_debug` setup into that function (so that the argument list contains `id, SEL, …` rather than `ConstPtr<objc_debug>, SEL, …`), you could then call `HostIMP::args_type_id` from it, I think?

Now that I think about it, adding a bound on `HostIMP` to the `msg_send` family would fix the problem where `msg_send` can be called with incorrect types for the first two arguments (should be `id` and `SEL`). But that's something I should do later.

src/frameworks/foundation.rs
hikari-no-yume: Ah, it's not accidental, but it could be removed if my suggestion below can be implemented: https://github.com/hikari-no-yume/touchHLE/pull/154/files#r1321731604.

src/objc/messages.rs
hikari-no-yume: Actually, if this is moved into `msg_send_debug`, I think it would be possible to do away with `objc_debug` and the two ABIs. You could put the type info in a special field of `ObjC` just before calling `call_from_host`, and then [take](https://doc.rust-lang.org/std/option/enum.Option.html#method.take) it inside `objc_msgSend_inner`. This ought to be clean and fast enough that it could even be done in release builds.

hikari-no-yume: Actually, it would even be possible to do away with `msg_send_debug`, I think? :)

hikari-no-yume: As a proof of concept I added `println!("{}", std::any::type_name::<(R, P)>());` to `msg_send` and it worked. (`std::any::type_name` shouldn't actually be used because it's not guaranteed to be unique, but might be useful for giving a nice error message.)

I'm realising this would simplify the patch so much that I could very quickly reimplement it myself from scratch. Uhh… how would you feel about that? ^^;

abnormalmaps: > The types being used to generate the type ID are not the actual types of the arguments, but the results of cloning them. So something weird could happen if there's references involved, maybe?

That's fair actually!  I actually originally also had CGPoint as Copy, but I ended up switching it. It's probably better that way.

> Actually, if this is moved into `msg_send_debug`, I think it would be possible to do away with `objc_debug` and the two ABIs. You could put the type info in a special field of `ObjC` just before calling call_from_host, and then take it inside `objc_msgSend_inner`. This ought to be clean and fast enough that it could even be done in release builds.

I thought about this and didn't really love the idea, but if it works for you then I'm OK with it :D 

Feel free to reimplement the patch!

hikari-no-yume:
I ended up implementing [my alternative approach](https://github.com/hikari-no-yume/touchHLE/pull/154#discussion_r1321731604) in 1bba0069e36ef62fa6ffd77e34e76356773d9500 instead.

Pros:

* Much smaller number of lines added overall.
* Also checks return types.
* More type safe. I even was able to ensure the first two args are `id` and `SEL` (392e3fac5516dff4b555540a3a0fede096f4899c).
* Only requires `'static`, not `Clone`.
* No ABI changes. Pushing more stuff onto the guest stack scares me! touchHLE is kinda designed around minimising the amount of stuff that has to care about the ABI…
* Better error message (I used `std::any::type_name`).
* No hashing: the type check is exact.
* I feel safe enabling it in release builds too.

Cons:

* Passing the type info via a new field on `ObjC` is ugly in its own way. In theory, there's a risk that the value wouldn't be cleared between message sends. But thankfully there's very few places that will actually need to touch it, so I think that is not going to be a problem in practice.

PR #160, opened 2023-09-24
Add hyperlinks in README.md

Vipitis:
add links to the markdown files using absolute paths


hikari-no-yume:
These are intentionally not linked because the links would be broken when it's included in releases or on the website. I guess I could fix that but I don't really want to, there's no value to the hyperlinks there IMO.

PR #186, opened 2023-11-01
Adds --y-tilt-offset=24 to KataDama options

nighto:
Fixes an issue on PC with controller that it would walk forward when left stick is untouched and not walk backwards when left stick is pressed down.

Before fix video: https://youtu.be/yjJbecDOXgs
After fix video: https://youtu.be/V_zpKfL6snk

Change-Id: Id46e75ad6bf0e3343d6fc135f0cbef86eacd5f28


Jaffhj:
nighto:fix-katadama

nighto:
Done on 8a0ac45f388c1e33e1c56ced358ff7070919bb39

PR #195, opened 2023-11-05
Bomberman Boots

TylerJaacks:
This isn't fully complete it is a lot of hacks to bypass function but it allows for Bomberman to boot. There are some functions in here that are complete. Also I would like some help with the CF* classes, as well if I could upstream some of my changes and improve my hacks. It should boot to the loading screen the reason that it doesn't move past that I assume has to do with CFRunLoop* related code not being full complete.

### DIFF
--- a/src/frameworks/foundation/ns_data.rs
+++ b/src/frameworks/foundation/ns_data.rs
@@ -146,6 +146,19 @@ pub const CLASSES: ClassExports = objc_classes! {
     );
 }
 
+-(()) getBytes:(MutPtr<u8>)buffer {
+    let host_object = env.objc.borrow::<NSDataHostObject>(this);
+
+    let size = host_object.length;
+    let bytes = host_object.bytes;
+
+    env.mem.memmove(
+        buffer.cast(),
+        bytes.cast_const(),
+        size
+    );
+}
+
 @end
 
 };
--- a/src/frameworks/foundation/ns_file_manager.rs
+++ b/src/frameworks/foundation/ns_file_manager.rs
@@ -5,8 +5,9 @@
  */
 //! `NSFileManager` etc.
 
+use std::ptr::null;
 use crate::mem::ConstPtr;
-use super::{ns_array, ns_string, NSUInteger};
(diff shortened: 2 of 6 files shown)


src/frameworks/foundation/ns_file_manager.rs
```diff
@@ -142,6 +145,45 @@ pub const CLASSES: ClassExports = objc_classes! {
     }
 }
 
+- (bool) changeCurrentDirectoryPath:(id)path {
```
TylerJaacks, line 148: - (bool) changeCurrentDirectoryPath:(id)path should be complete.

src/frameworks/foundation/ns_file_manager.rs
```diff
@@ -142,6 +145,45 @@ pub const CLASSES: ClassExports = objc_classes! {
...
+            log!("Warning: chdir({:?}) failed, could not change working directory to {:?}, returning -1", path_str, path);
+
+            return false;
+        }
+    }
+
+    return true;
+}
+
+- (id) currentDirectoryPath {
```
TylerJaacks, line 171: - (id) currentDirectoryPath should be complete.

src/frameworks/foundation/ns_data.rs
```diff
@@ -146,6 +146,19 @@ pub const CLASSES: ClassExports = objc_classes! {
     );
 }
 
+-(()) getBytes:(MutPtr<u8>)buffer {
```
TylerJaacks, line 149: -(()) getBytes:(MutPtr<u8>)buffer this function should be complete.

ciciplusplus:
Hello @TylerJaacks,

Please follow this guide https://github.com/hikari-no-yume/touchHLE/blob/trunk/CONTRIBUTING.md if you want to contribute.

In particular:
- you should push your commits to Gerrithub with Change-Ids
- please abide to https://github.com/hikari-no-yume/touchHLE/blob/trunk/dev-docs/code-style.md
- please separate your PR to smaller commits. For example, you should separate complete implementation from incomplete one
- please remove profanities from commit messages 👼 

Thanks

PR #205, opened 2023-11-23
Jackpot Casino default options

nighto:
Change-Id: Iee9ad43889d46758bb2400422ca03a2e33dafd0b


ciciplusplus:
Just FYI, i think for the pure config change like this one, you don't need to create Github PRs, Gerrit patchsets should be sufficient.

nighto:
Gotcha
Fixed with 24689c512b39c266e22de770d4ac944ea045053f

PR #206, opened 2023-11-23
MotoRacing GP default options

nighto:
- adds --landscape-right
- adds controller mappings

Change-Id: Iea0664f529281913db090d7ce71b7d8989e7752d


nighto:
Closed this one and only created gerrit patch: https://review.gerrithub.io/c/hikari-no-yume/touchHLE/+/1176050

PR #231, opened 2024-01-10
Earthworm Jim

ciciplusplus:
### Tested versions
- 1.0.8

### Known issues
- In-game splash screens are not properly rendering "dim" state
- Level select screen have rendering issues
- Music support is in WIP state

Note: this is on top of fifa-10 branch/PR

### Screenshots

Screenshot 2024-01-11 at 00 16 56
Screenshot 2024-01-11 at 00 17 19

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,6 +30,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2cb2f989d18dd141ab8ae82f64d1a8cdd37e0840f73a406896cf5e99502fab61"
 
 [[package]]
+name = "arrayvec"
+version = "0.7.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
+
+[[package]]
 name = "atty"
 version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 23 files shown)


omagadsamsa:
Another bug: Japanese doesn't work, English replaces it

omagadsamsa:
and after some time spent in the menu with the Japanese application crashes

omagadsamsa:
even with the Japanese language when you are already playing you kill a cow and then you hit the tires and climb on the excavator and it crashes

ciciplusplus:
> and after some time spent in the menu with the Japanese application crashes

do you have a crash stack?

omagadsamsa:
> do you have a crash stack?

I don’t know what it is but I’ll record a video of how it happens

omagadsamsa:
after departure from the menu, I recommend rewinding to the end of the video, there is a departure https://youtu.be/zD9XtBqtsVo

ciciplusplus:
Thanks. The crash stack is a last lines of text printed in your command line window, usually starting with something like `thread 'main' panicked at ...` (next time you can just copy past it instead of recording the whole video).

I fixed the crash, but there is still an issue with background audio - tracks played only once instead of being played in the loop.

omagadsamsa:
> Thanks. The crash stack is a last lines of text printed in your command line window, usually starting with something like `thread 'main' panicked at ...` (next time you can just copy past it instead of recording the whole video).
> 
> I fixed the crash, but there is still an issue with background audio - tracks played only once instead of being played in the loop.

Okay, thanks for seeing the app problem and fixing the crash!

ciciplusplus:
Mostly merged, the game is now playable on the trunk. Background music is not supported as of now.
Closing

PR #347, opened 2024-08-05
Fix Android CI build

ciciplusplus:
Github runner seems to bump a vendored version of NDK to version 27, resulting in an incorrect value of ANDROID_NDK_ROOT being set.

The problem is fixed by explicitly setting that variable with a path to NDK with a correct version (the same version as in build.gradle file "25.2.9519653")


ciciplusplus:
Merged as https://github.com/touchHLE/touchHLE/commit/dcef0c0218253ddf7522d517649adf5d69884b57

PR #351, opened 2024-08-22
Fix Android CI build (again)

ciciplusplus:
CI runner doesn't include anymore the version of Android NDK which we're using, so we need to install it manually

Change-Id: I92749c107ca9ff45053454db5f8043b1924e286c


ciciplusplus:
Merged as https://github.com/touchHLE/touchHLE/commit/ba07bd31000edf1ff4e9f93e1d15d7c43c95b7a0

PR #455, opened 2025-05-21
Update readme ios support

htthanhvn:


ciciplusplus:
WTF is this PR

hikari-no-yume:
Please do not use AI models/LLMs/ChatGPT to make PRs. We cannot accept them for legal reasons, and these systems do not understand touchHLE anyway, so it is just a waste of time.

PR #457, opened 2025-05-27
Fix NPOT/PowerVR texture bug

chyyran:
Fixes NPOT texture rendering

* For textures loaded from image data, set `CLAMP_TO_EDGE`
* For render buffer, set up a POT texture to copy the guest framebuffer into, then properly handle texture coordinates to only render from written parts of the POT texture.


chyyran:
Oops, looks like there's a rebase artifact leftover. Let me fix that...

Issue #500, opened 2025-09-20
8-directional D-pad support for button-to-touch arguments


vaguerant:
### Describe the feature or issue.

Currently, the `button-to-touch` argument simply treats the D-pad as four buttons. While this is accurate to their real-world functionality, in practice their behavior is more akin to eight buttons, with the four cardinal directions *and* the four diagonals, where two buttons are pressed simultaneously.

iPhone virtual D-pads such as the one used in Earthworm Jim reflect this behavior: the on-screen D-pad is actually made up of eight different touch zones. To make the character shoot in eight directions, all eight touch points need to be reachable. With the current 4-way setup, it is impossible to aim/shoot diagonally when using a controller: multi-touching two cardinal direction buttons is *not* registered as a diagonal input by the game.

I will propose two possible solutions, with their own pros and cons. If you have a different solution in mind, feel free to ignore these.

**Solution 1: More `button-to-touch` values**
Add `DPadUpLeft`, `DPadUpRight`, `DPadDownLeft` and `DPadDownRight` as additional values for the `button-to-touch` argument. If arguments with these values are declared, treat the diagonals as unique individual button presses. i.e. When `DPadUp` and `DPadRight` are registering simultaneously **and** `DPadUpRight` is declared as a `button-to-touch` argument, do **not** evaluate the cardinal values. **Only** evaluate the virtual touch set for `DPadUpRight`.

Conversely, if the corresponding diagonal has not been declared in an argument, evaluate the D-pad presses as they are now: two multi-touch presses at different points. This should preserve compatibility in case of titles which *do* support the multi-touch approach.

Pros:
- extremely precise manual diagonal touch point positioning
- seems easier to implement

Cons:
- double the setup labor (whoever makes the config needs to find 8 touch points)
- considerable argument bloat (double the D-pad arguments)

**Solution 2: Add an `8way-DPad` argument**
With this approach, 4-way input as it exists now would work the same, but if the argument `--8way-DPad` (or similar) is passed, touchHLE would interpolate the values of any two virtual presses assigned to the D-pad and touch the point directly in-between those two points instead of either individual point.

e.g. If `DPadUp` is assigned to 40,280 and `DPadRight` is assigned to 60,300, pressing both at the same time will instead register a touch at 50,290, averaging the respective *x* and *y* offsets. Obviously, this would require some rounding in practice.

Pros:
- only 4 reference touch points needed (reduced labor/config bloat)

Cons:
[...]

Issue #400, opened 2024-12-08
Map taps of any coordinate to inputs for keyboard and enhancement


dante3732:
### Describe the feature or issue.

Good afternoon, some games when using the mouse it is difficult to play only on it, some use character movement on the virtual stick, it would be good to use the keyboard for walking and other functions.

I also noticed that not all buttons from the gamepad can be assigned to buttons, it would be nice to add all the available ones.

Stick emulation is also needed, as normal stick edge pressing doesn't work for walking. You can test this in Zombie Infection. 
You can still add a modifier on the keyboard for the stick. For example, if you use the WSAD buttons to walk slowly, and if you press the shift button, the character can move faster.
You can use this feature in Android emulators Nox or Bluestacks for inspiration. 

### By submitting this issue, I certify that…

- [X] I understand that linking to, discussing or asking for pirated content is not permitted, and will lead to an immediate ban.
- [X] I understand that “Abandonware” and illegal copies of free games are also pirated content.

999pingGG:
I second this! Make keyboard mappings work just like joystick button mappings

Issue #325, opened 2024-06-24
Rayman 2 controller settings


idontwannaforgetmyusernameorpassword:
The default settings don't have --button-to-touch settings for Rayman 2. Anyone got some?

ciciplusplus:
This is not as straightforward as it may seems, as we don't have support for simulating "pull-to-move" thumbsticks required for Rayman 2.

ciciplusplus:
Fixed with https://github.com/touchHLE/touchHLE/commit/542dae74037851c8a6d5424efdd12e5dc99771e8

PR #513, opened 2025-11-20
Add input handling for analog stick, DPad

celerizer:
This PR adds two new options to improve game input on controllers:

**--stick-to-touch=x,y,w,h** defines a region on the screen in which a touch with FingerId::StickToTouch slides around relative to the left analog stick. This is useful for games like Resident Evil 4 mobile edition and Rayman 2 which use analog movement, which were previously very stiff to control.

**--dpad-to-touch=x,y,w,h** defines a region on the screen in which a touch with FingerId::DpadToTouch mimics an 8-directional pad. This is useful for 2D games where diagonals are used.

In implementing these I added a little bit of controller state to the Window class. I also removed a check in **ui_touch.rs** that makes sure only one touch is happening at a time, which I don't think is needed anymore?

I'm opening this PR here for thoughts on the implementation and for others to test. A few quirks still exist such as how this should play with the accelerometer that's already mapped to the left stick, and I don't have access to any games that would benefit from DPad like Earthworm Jim. I can open on Gerrit if there are no major problems/disagreements with adding this.

https://github.com/user-attachments/assets/e477fc60-69b8-4feb-b5d2-edb20a49feb9

image

Related: #325 #400 #500

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -523,37 +523,16 @@ impl Window {
                         && options.dpad_to_touch.is_some()
                     {
                         let Some((x, y, w, h)) = options.dpad_to_touch else {
-                            continue;
+                            unreachable!();
                         };
+
                         // Update held state
-                        match (button, &event) {
-                            (crate::options::Button::DPadLeft, E::ControllerButtonDown { .. }) => {
-                                self.dpad_state.left = true
-                            }
-                            (crate::options::Button::DPadLeft, E::ControllerButtonUp { .. }) => {
-                                self.dpad_state.left = false
-                            }
-
-                            (crate::options::Button::DPadRight, E::ControllerButtonDown { .. }) => {
-                                self.dpad_state.right = true
-                            }
-                            (crate::options::Button::DPadRight, E::ControllerButtonUp { .. }) => {
-                                self.dpad_state.right = false
-                            }
-
-                            (crate::options::Button::DPadUp, E::ControllerButtonDown { .. }) => {
-                                self.dpad_state.up = true
-                            }
-                            (crate::options::Button::DPadUp, E::ControllerButtonUp { .. }) => {
-                                self.dpad_state.up = false
-                            }
-
-                            (crate::options::Button::DPadDown, E::ControllerButtonDown { .. }) => {
-                                self.dpad_state.down = true
-                            }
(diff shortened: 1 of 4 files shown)


src/window.rs
```diff
@@ -495,30 +515,152 @@ impl Window {
...
-                            )]))
+                    // Called whenever a DPad direction is pressed or released
+                    if (button == crate::options::Button::DPadLeft
+                        || button == crate::options::Button::DPadUp
+                        || button == crate::options::Button::DPadRight
+                        || button == crate::options::Button::DPadDown)
+                        && options.dpad_to_touch.is_some()
+                    {
+                        let Some((x, y, w, h)) = options.dpad_to_touch else {
+                            continue;
```
hikari-no-yume: `unreachable!()` would be more appropriate here.

hikari-no-yume:
Hi! Thanks for this contribution, this is a long-overdue feature and a great enhancement to have. I think I would like to merge this into the next release (which will hopefully be out before the end of the year...)

One concern I have is that there's only two sticks available on a gamepad, and we already have assigned meanings to two of them: one controls the accelerometer tilt simulation, and the other controls the virtual cursor. The new `--stick-to-touch=` would replace the former, but there might be games which use both tilt and an analog stick, or which even have two sticks. If it would be possible to specify which analog stick is used as input, that might make for a nice addition, but perhaps that's too complicated.

In any case: please do squash this into a single commit and push to Gerrit for further review. :)

src/window.rs
celerizer: True, true. Done

celerizer:
I don't know of any games that use accel and a virtual stick at the same time. The only thing I know of that uses both is Rayman 2, but the accel and platforming segments are separated, so having both tied to the left stick doesn't cause an issue. That particular game plays seamlessly with both on the same stick but I agree something else might not.

Adding right stick seems a little iffy right now because the virtual pointer is basically always a useful feature. Maybe something to look into if more games with more complex controls start being supported? I don't know the best way to handle it either

I opened this on Gerrit just now

ciciplusplus:
Merged as https://github.com/touchHLE/touchHLE/commit/542dae74037851c8a6d5424efdd12e5dc99771e8 with small changes. Many thanks for your contribution!

celerizer:
Cool stuff, Merry Christmas

PR #536, opened 2026-01-17
Refactor GitHub Actions workflow for touchHLE

tarqsbay74-png:


ciciplusplus:
wtf

hikari-no-yume:
I can save you the effort: this change would not produce a 32-bit build of touchHLE for Android that actually works.

PR #541, opened 2026-01-27
add support for importing IPA files via any external file manager on …

nerivalaitis:
Makes it easier to import IPA files into touchHLE. This is especially helpful on newer Android devices, where file access is more restricted and moving files between public and private storage areas can be tricky.


hikari-no-yume:
Hi! Thanks for your contribution.

I don't think it would be a good idea for us to add something which allows importing files but not exporting them. That risks creating a class of users who can't back up their save data or edit the settings.

PR #597, opened 2026-04-09
Refactor NSBundle implementation for resource paths

FelipeVidal2022:


hikari-no-yume:
This PR makes no sense and probably doesn't even compile. Please test things *before* pushing.

PR #602, opened 2026-04-12
Remove symlinks from android build.

acieslewicz:
Change-Id: I5f7c73d763e199c133c9dfc9f66e91b113b7a5cd


acieslewicz:
We can definitely remove the SDL symlinks if the custom stack size is not needed in SDLActivity. Im not so sure about removing the src symlink. For some reason it sometimes works locally and sometimes doesn't (caching?). Wanted to test in CI since that should be clean.

hikari-no-yume:
Definitely would prefer it if we can go back to just symlinking everything. The Gerrit review UI handles that one copied-and-modified java file _horribly_.

android/app/build.gradle.kts
```diff
@@ -102,6 +102,12 @@ android {
         }
     }
 
+    sourceSets {
+        getByName("main") {
+            java.srcDir("${rootDir.parentFile}/vendor/SDL/android-project/app/src/main/java")
```
hikari-no-yume, line 110: Ohh, you just change the build definition to reference the vendor directory, rather than using a symlink? Is that what this is doing? That's very nice, if so.

acieslewicz: Yep, that's exactly it!

PR #626, opened 2026-06-08
Tilt to live hd

Ahmadjerj:
#  Tilt to Live HD & optimizations

Hey guys, so i worked like 2 weeks on this. i was working mainly on the HD version of tilt to live but it happened to support the non hd version too which is nice.

also i had done some optimization I saw in my way, like speeding up the dyld host lookup and expanding some libc stuff so it doesn't crash as much. 

I tried to follow the rules and formatting at the last minute because honestly i wasn't gonna submit the code originally but decided to share it. 

I  probably won't be active on gerrithub if anything gets rejected for any reason so if anything gets rejected or I am needed to fix something, just let me know here on github or by email
### Previews

image
image


hikari-no-yume:
This is a huge amount of code. Was an AI assistant involved in writing it?

hikari-no-yume:
I guess on balance it doesn't seem like AI-generated code actually, though I would still like to know.

Ahmadjerj:
> This is a huge amount of code. Was an AI assistant involved in writing it?

I didn't use Ai for writing the code so this isn't code generated by Ai but I did use AI for 1 thing
I hadn't followed the naming style in the changes I had done so I gave an agent a task to do that

And I just had noticed the Ai policy that forbiddes the use of it, I am so sorry I thought it was just for code not in general use case

If the PR gets rejected because of that, I understand 

Also the glkit was my attempt to get the ios 6 build working before I had found the 3.2 ios build

PR #630, opened 2026-06-11
split deadzone setting into left and right

kylon:
it is not unusual to have one stick with more drift than the other

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -682,7 +682,7 @@ impl Window {
                     if axis == sdl2::controller::Axis::LeftX
                         || axis == sdl2::controller::Axis::LeftY
                     {
-                        let (stick_x, stick_y, _) = self.get_controller_stick(options, true);
+                        let (stick_x, stick_y, _) = self.get_controller_stick(options.left_deadzone, true);
                         let coords = transform_input_coords(
                             self,
                             (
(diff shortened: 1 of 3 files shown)


src/window.rs
```diff
@@ -1108,7 +1108,7 @@ impl Window {
     /// Get the summed X and Y positions and button state of the left or right
     /// analog stick of the game controllers. Each axis value is in the range
     /// [-1, 1].
-    fn get_controller_stick(&self, options: &Options, left: bool) -> (f32, f32, bool) {
+    fn get_controller_stick(&self, deadzone: f32, left: bool) -> (f32, f32, bool) {
```
hikari-no-yume, line 1111: Since this takes a `bool` which selects which stick to use, wouldn't it make more sense for it to select the correct deadzone from the options itself? Otherwise it would be possible to accidentally call it with the wrong deadzone, and I'm not sure why that would be desirable.

hikari-no-yume (changes requested):

hikari-no-yume:
You typo'd the commit message and PR title btw.

src/window.rs
kylon: i just began with rust, i thought that passing a float, instead of ref to options, just for one value, may be more efficient? basic data type and no additional if, because in all calls you already know what stick it is, due to the bool arg

PR #633, opened 2026-06-14
Update ui_web_view.rs

fuukitoyama:
Add stub for instanceMethodSignatureForSelector to prevent crash in Fruit Ninja 1.3

### DIFF
--- a/src/frameworks/uikit/ui_view/ui_web_view.rs
+++ b/src/frameworks/uikit/ui_view/ui_web_view.rs
@@ -27,6 +27,12 @@ pub const CLASSES: ClassExports = objc_classes! {
 - (())setDelegate:(id)_delegate {
     // TODO
 }
+    - (id) instanceMethodSignatureForSelector:(id) _sel {
+        // Return nil to inform the Objective-C runtime that no signature exists.
+        // This prevents the emulator from crashing and bypasses the web request.
+        nil
+    }
+
 - (())loadRequest:(id)request { // NSURLRequest*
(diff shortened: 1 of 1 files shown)


src/frameworks/uikit/ui_view/ui_web_view.rs
```diff
@@ -27,6 +27,12 @@ pub const CLASSES: ClassExports = objc_classes! {
 - (())setDelegate:(id)_delegate {
     // TODO
 }
+    - (id) instanceMethodSignatureForSelector:(id) _sel {
```
hikari-no-yume, line 30: - This doesn't make sense: this is supposed to be a class method, but you have defined an instance method.
- This isn't specific to `UIWebView` and shouldn't be defined here, it is a generic `UIObject` method.
- The formatting here doesn't match the surrounding code.

src/frameworks/uikit/ui_view/ui_web_view.rs
```diff
@@ -27,6 +27,12 @@ pub const CLASSES: ClassExports = objc_classes! {
...
+        // Return nil to inform the Objective-C runtime that no signature exists.
+        // This prevents the emulator from crashing and bypasses the web request.
```
hikari-no-yume, line 32: This doesn't inform the Objective-C runtime of anything, the caller of this function is elsewhere.

hikari-no-yume (changes requested):
We don't accept patches as GitHub PRs, they should be submitted to Gerrit.

